PR #4, opened 2020-03-21
Stop namespacing the pages inside `coronavirus-form`

issyl0:
A) This is the vulnerable people form, we've made it a separate app.
B) It's simpler this way.


issyl0:
I've decided it's not worth it. We'll just stick with it how it is to start on the questions, as there's a bunch of other things that expect `CoronavirusForm`, and until the business form people refactor theirs is good to be consistent.

PR #8, opened 2020-03-21
Q2 Medical Conditions page

issyl0:
Screenshot 2020-03-21 at 14 40 20

This links to the "have you been tested?" page that's the next question.

https://trello.com/c/k8Z8x4Hc/26-do-you-have-any-medical-conditions-page

### DIFF
--- a/app/views/coronavirus_form/medical_conditions.html.erb
+++ b/app/views/coronavirus_form/medical_conditions.html.erb
@@ -16,7 +16,7 @@
       heading: t('coronavirus_form.medical_conditions.title'),
       is_page_heading: true,
       name: "medical_conditions",
-      description: t('coronavirus_form.medical_conditions.description').html_safe,
+      description: sanitize(t('coronavirus_form.medical_conditions.description')),
       items: t('coronavirus_form.medical_conditions.options').map do |_, item|
         {
           value: item[:label],
--- a/app/views/coronavirus_form/medical_conditions.html.erb
+++ b/app/views/coronavirus_form/medical_conditions.html.erb
@@ -1,0 +1,30 @@
+<% content_for :title do %><%= t('coronavirus_form.medical_conditions.title') %><% end %>
+    <% content_for :meta_tags do %>
+      <meta name="description" content="<%= t('coronavirus_form.medical_conditions.title') %>" />
+    <% end %>
+    
+    <% content_for :back_link do %>
+      <%= render "govuk_publishing_components/components/back_link", { href: previous_path } %>
+    <% end %>
+    
+    <%= form_tag({},
+      "data-module": "track-coronavirus-form-vulnerable-people-medical_conditions",
+      "data-question-key": "medical_conditions",
+      "id": "medical_conditions"
+    ) do %>
+    <%= render "govuk_publishing_components/components/radio", {
+      heading: t('coronavirus_form.medical_conditions.title'),
+      is_page_heading: true,
+      name: "medical_conditions",
+      description: t('coronavirus_form.medical_conditions.description').html_safe,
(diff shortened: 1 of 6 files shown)


app/views/coronavirus_form/medical_conditions.html.erb
```diff
@@ -0,0 +1,30 @@
...
+    <%= form_tag({},
+      "data-module": "track-coronavirus-form-vulnerable-people-medical_conditions",
+      "data-question-key": "medical_conditions",
+      "id": "medical_conditions"
+    ) do %>
+    <%= render "govuk_publishing_components/components/radio", {
+      heading: t('coronavirus_form.medical_conditions.title'),
+      is_page_heading: true,
+      name: "medical_conditions",
+      description: t('coronavirus_form.medical_conditions.description').html_safe,
```
injms: Sorry, I should've suggested `sanitize()` instead of `.html_safe`:

```suggestion
      description: sanitize(t('coronavirus_form.medical_conditions.description')),
```

config/locales/en.yml
```diff
@@ -28,6 +28,27 @@ en:
           label: "No"
         not_sure:
           label: "Not sure"
+    medical_conditions:
+      title: Do you have a medical condition that makes you vulnerable to coronavirus?
+      description: |
+        You’re vulnerable if you:
+        <ul>
```
injms: I think we should use the Design System styles for the list:

```suggestion
        <ul class="govuk-list govuk-list--bullet"> 
```

injms (changes requested):

config/locales/en.yml
```diff
@@ -28,6 +28,27 @@ en:
           label: "No"
         not_sure:
           label: "Not sure"
+    medical_conditions:
+      title: Do you have a medical condition that makes you vulnerable to coronavirus?
+      description: |
+        You’re vulnerable if you:
```
injms: Need to add the paragraph text class in to make sure there's spacing after the paragraph and before the list:

```suggestion
        <p class="govuk-body">You’re vulnerable if you:</p>
```

injms (changes requested):

injms (approved):
🚢

config/locales/en.yml
```diff
@@ -28,6 +28,27 @@ en:
...
+          <li>have had a solid organ transplant</li>
+          <li>have lung cancer, and you’re getting active chemotherapy or radical radiotherapy</li>
+          <li>have cancer of the blood or bone marrow, at any stage of treatment - for example, leukaemia, lymphoma or myeloma</li>
+          <li>have any cancer for which you’re getting immunotherapy or other continuing antibody treatments</li>
+          <li>have any cancer for which you’re getting a targeted treatment which can affect the immune system - for example, protein kinase inhibitors or PARP inhibitors</li>
+          <li>have had bone marrow or stem cell transplants in the last 6 months, or are still taking immunosuppression drugs</li>
+          <li>have a severe respiratory condition - including cystic fibrosis, severe asthma or severe COPD (Chronic Obstructive Pulmonary Disease)</li>
+          <li>have a rare disease or inborn error of metabolism that significantly increases your risk of infection - for example SCID or homozygous sickle cell</li>
+          <li>are getting an immunosuppression therapy</li>
+          <li>are pregnant and have a significant congenital heart disease</li>
```
bilbof, line 45: ~I wonder if this could go into translations~ it already is 🤦‍♂

bilbof: Though perhaps we could keep the HTML tags in the templates, and just have the raw text in here.

bilbof (approved):
LGTM 🚢

PR #10, opened 2020-03-21
Add "Do you have a history of addiction?" question

brucebolt:
Adds the "Do you have a history of addiction?" question, with placeholder next and previous locations.

Screenshot 2020-03-21 at 15 10 37

Trello card: https://trello.com/c/gxj9zkcc

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -32,7 +32,7 @@ en:
       title: Do you have a medical condition that makes you vulnerable to coronavirus?
       description: |
         <p class="govuk-body">You’re vulnerable if you:</p>
-        <ul class="govuk-list govuk-list--bullet"> 
+        <ul class="govuk-list govuk-list--bullet">
           <li>have had a solid organ transplant</li>
           <li>have lung cancer, and you’re getting active chemotherapy or radical radiotherapy</li>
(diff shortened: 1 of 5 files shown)


config/locales/en.yml
```diff
@@ -37,3 +37,10 @@ en:
           label: Yes, I've been tested and I do not have coronavirus
         option_no:
           label: No, I have not been tested
+    addiction:
+      title: Do you have a history of addiction?
+      options:
+        option_yes:
+          label: Yes, I’ve had problems with drug addiction or alcohol addiction (alcoholism)
+        option_no:
+          label: "No"
```
issyl0, line 67: Can we be consistent about the use of quotes or not in these options? Some have them, some don't, this one is inconsistent in itself...

issyl0 (commented):
LGTM. One YAML style nitpick but it shouldn't block shipping this, it's just potentially confusing to devs.

issyl0 (approved):
And I meant the above to be an approval...

config/locales/en.yml
brucebolt: My preference is to not have them, but "No" is a reserved word, so needs the quotes.  Maybe we should just quote everything?

PR #14, opened 2020-03-21
Add "Do you know your NHS number?" question

issyl0:
No spacing:

Screenshot 2020-03-21 at 15 52 41

With spacing (like prototype):

Screenshot 2020-03-21 at 16 06 11

Links to dummy pages before/after.

### DIFF
--- a/app/controllers/coronavirus_form/know_nhs_number_controller.rb
+++ b/app/controllers/coronavirus_form/know_nhs_number_controller.rb
@@ -1,0 +1,40 @@
+# frozen_string_literal: true
+
+class CoronavirusForm::KnowNhsNumberController < ApplicationController
+  include ActionView::Helpers::SanitizeHelper
+  include FieldValidationHelper
+
+  def show
+    render "coronavirus_form/#{PAGE}"
+  end
+
+  def submit
+    know_nhs_number = sanitize(params[:know_nhs_number]).presence
+    session[:know_nhs_number] = know_nhs_number
+
+    invalid_fields = validate_radio_field(
+      PAGE,
+      radio: know_nhs_number,
+    )
+
+    if invalid_fields.any?
+      flash.now[:validation] = invalid_fields
+      render "coronavirus_form/#{PAGE}"
+    elsif session["check_answers_seen"]
+      redirect_to controller: "coronavirus_form/check_answers", action: "show"
+    elsif session[:know_nhs_number] == I18n.t("coronavirus_form.know_nhs_number.options.option_no.label")
+      redirect_to controller: "coronavirus_form/essential_supplies", action: "show"
+    else
+      redirect_to controller: "coronavirus_form/#{NEXT_PAGE}", action: "show"
+    end
+  end
+
+private
+
+  PAGE = "know_nhs_number"
+  NEXT_PAGE = "check_answers"
+
+  def previous_path
+    coronavirus_form_contact_details_path
+  end
+end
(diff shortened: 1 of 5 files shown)


leenagupte (dismissed):
Not a blocker, but there is supposed to be an empty line after the first line of the hint text.

issyl0:
Frontend devs (@injms @alex-ju) please tell me better ways of spacing paragraphs. If I made them actual paragraphs there was too much space between the title and the first line of the hint, so I hacked it with `<br />`. 🤢

issyl0:
Also this needs an `elsif` to account for skipping the (as yet unwritten) NHS number question if the user doesn't know their NHS number... I'll borrow from the business form...

config/locales/en.yml
```diff
@@ -73,6 +73,17 @@ en:
           label: "Yes, I’ve had problems with drug addiction or alcohol addiction (alcoholism)"
         option_no:
           label: "No"
+    know_nhs_number:
+      title: "Do you know your NHS number?"
+      hint: |
+        An NHS number is a 10 digit number, like 485 777 3456.
+        <br /><br />
```
alex-ju, line 115: This is rather atypical as we don't tend to have hints with multiple paragraphs. I think this should do for now.

issyl0: Thanks Alex!

issyl0:
> Also this needs an `elsif` to account for skipping the (as yet unwritten) NHS number question if the user doesn't know their NHS number... I'll borrow from the business form...

On this, [the business form has](https://github.com/alphagov/govuk-coronavirus-business-volunteer-form/blob/f53e993aaca4cf0a58d233d72bf4b0869c5d5c04/app/controllers/coronavirus_form/offer_transport_controller.rb#L25-L26):

```ruby
    elsif session[:offer_transport] == I18n.t("coronavirus_form.#{PAGE}.options.option_yes.label")
      redirect_to controller: "coronavirus_form/transport_type", action: "show"
```

But I can't figure out what the `yes` or `no` options are for this one. 'Cause we want to go to the next question after the "enter your NHS number" one if the user says no.

app/controllers/coronavirus_form/know_nhs_number_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+    end
+  end
+
+private
+
+  PAGE = "know_nhs_number"
+  NEXT_PAGE = "check_answers"
+
+  def previous_path
+    "/" # TODO
```
huwd:
```suggestion
   coronavirus_form_contact_details_path
```

I think this one has been merged now

config/locales/en.yml
```diff
@@ -108,6 +108,17 @@ en:
         label: "Phone number (for text messages)"
       email:
         label: "Email address"
+    know_nhs_number:
+      title: "Do you know your NHS number?"
+      hint: |
+        An NHS number is a 10 digit number, like 485 777 3456.
```
huwd, line 114: is spacing too big if we use `<p>` tags here?

huwd (approved):
One question and a suggestion, but looks great!

config/locales/en.yml
issyl0: Yes.

app/controllers/coronavirus_form/know_nhs_number_controller.rb
issyl0: Thanks - applied.

PR #15, opened 2020-03-21
Add Question 4: What is the address where you need support?

huwd:
# What
@ TODO
# Screenshots
**mobile**
image

**desktop**
image

### Validation

**mobile**
image

**desktop**
image

### DIFF
--- a/app/controllers/coronavirus_form/support_address_controller.rb
+++ b/app/controllers/coronavirus_form/support_address_controller.rb
@@ -1,0 +1,87 @@
+# frozen_string_literal: true
+
+class CoronavirusForm::SupportAddressController < ApplicationController
+  include ActionView::Helpers::SanitizeHelper
+  include FieldValidationHelper
+
+  REQUIRED_FIELDS = %w(
+      building_and_street_line_1
+      town_city
+  ).freeze
+
+  def show
+    session[:support_address] ||= {}
+    render "coronavirus_form/#{PAGE}"
+  end
+
+  def submit
+    session[:support_address] ||= {}
+    session[:support_address]["building_and_street_line_1"] = sanitize(params[:building_and_street_line_1]).presence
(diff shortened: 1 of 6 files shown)


app/controllers/coronavirus_form/support_address_controller.rb
```diff
@@ -0,0 +1,87 @@
...
+
+  def validate_conditionally_present_fields(product)
+    return [] if product["postcode"].present? || product["county"].present?
+
+    invalid_fields = []
+
+    if product["postcode"].blank?
+      invalid_fields << {
+        field: "postcode",
+        text: t("coronavirus_form.errors.missing_postcode_or_county_fields"),
```
issyl0: Given we have both postcode and county fields, this two-in-one error message feels confusing. 🤔

config/locales/en.yml
```diff
@@ -123,3 +124,16 @@ en:
         label: Middle names (optional)
       last_name:
         label: Last name
+    support_address:
+      title: What is the address where you need support?
+      hint: If you're applying for someone else, enter the address of the person who needs support.
```
```suggestion
      hint: If you’re applying for someone else, enter the address of the person who needs support.
```

app/controllers/coronavirus_form/support_address_controller.rb
huwd: ah i see, like keep the message but split it out?
I'm still waiting on final copy - but the deal is you must have at least one of them

huwd: i've amended that to split them out, hopefully a bit less confusing?

huwd:
Also. thanks to issy's suggestion, noticed the hint text wasn't rendering (see the screenshots).
That's now fixed

bilbof (approved):

issyl0 (approved):
Thanks Huw!

PR #21, opened 2020-03-21
Add pry and awesome print

huwd:
# What

Switch to using pry for rails console and allows for the debugger

# Why

I generally find pry easier to work with when dropping in a debugger, pry rails makes pry the default for `rails c` and awesome_print gives you nicely coloured and formatted outputs for things like hashes.

I'm about to muck around with some gnarly validation and feel like i might need to do some Pry Driven Development :D

Feel free to shoot this down if the majority's workflow prefers IRB?


richardTowers:
See https://github.com/alphagov/govuk-coronavirus-business-volunteer-form/pull/46 for a fix for your travis issues :)

huwd:
@richardTowers: gentleman and a scholar

richardTowers (approved):

PR #25, opened 2020-03-21
rubocop: Exclude config/routes.rb from Metrics/BlockLength

issyl0:
- When I rebased one of the question PRs, RuboCop moaned about this.
- There are going to be many more routes and we can't help that...


leenagupte (approved):
💯

PR #26, opened 2020-03-21
Check for session data for viewing the check your answers page.

theseanything:
This is very blunt and simple check to see if the user has at least
visited the first question before allowing them to view the check
your answers page. This should help prevent incorrect sumbission for
people who've not completed the full question flow.


bevanloon (approved):

bilbof (approved):
Cool 👍 we can add to that if statement if we want to check more

PR #28, opened 2020-03-21
Add exclude for rubocop block length rule

huwd:
#Why

Our routes get loooooonnng for this app. and the linting doesn't provide any value here. We can use code-review to keep each other honest about complexity


issyl0:
Already done in #25.

PR #32, opened 2020-03-21
Make validation errors appear on all pages when needed

issyl0:
- We missed including `error_message` in the component, so it didn't
  know which error message to render.
- This also sets snappier error messages than re-printing the question
  titles, for those that didn't already have them.

A selection of screenshots is below...

Screenshot 2020-03-21 at 19 54 30

Screenshot 2020-03-21 at 19 54 59

Screenshot 2020-03-21 at 19 54 22

### DIFF
--- a/app/views/coronavirus_form/know_nhs_number.html.erb
+++ b/app/views/coronavirus_form/know_nhs_number.html.erb
@@ -16,6 +16,7 @@
       heading: t('coronavirus_form.know_nhs_number.title'),
       is_page_heading: true,
       name: "know_nhs_number",
+      error_message: error_items('know_nhs_number'),
       hint: sanitize(t('coronavirus_form.know_nhs_number.hint')),
       items: t('coronavirus_form.know_nhs_number.options').map do |_, item|
         {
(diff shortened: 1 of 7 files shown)


app/views/coronavirus_form/know_nhs_number.html.erb
```diff
@@ -8,14 +8,15 @@
     <% end %>
     
     <%= form_tag({},
-      "data-module": "track-coronavirus-form-vulnerable-people-know_nhs_number",
+      "data-module": "track-coronavirus-form-vulnerable-people-know-nhs-number",
```
issyl0: For tomorrow: I don't think this is needed, it's got underscores in all the other places...

leenagupte: You're right this isn't needed 😄

leenagupte (approved):
This looks great! Thanks for fixing this!

issyl0:
I don't know if I got all the things, but when we do the flow and some light QA we can fix the others. I also don't know if all the questions I fixed are still needed. But again, when we do the flow and QA...

PR #33, opened 2020-03-22
Use log-formatter to log to JSON

richardTowers:
This should make the logs more useful for Splunk etc.

I've also removed the modernizr include, because it was 404ing and
giving me annoying 404 logs. As far as I can tell we don't actually use
the modernizr JS anywhere (I might be missing some CSS rule that uses
the classes it writes on body though).


richardTowers:
Not sure about adding this dependency:

https://github.com/chadlwm/log_formatter

It seems to work, but it's not a popular as say "lograge".

OllieJC (commented):
:raised_hands:

bilbof (approved):

PR #35, opened 2020-03-22
Add annotations for routes

huwd:
Thought this would be useful, general idea.
- Annotate the version of the prototype we're on `(v3)`
- Leave a comment with nothing below it if we spot a question that we don't have a trello ticket for
- Comment out removed routes at the bottom until they're confirmed and we can delete them.


bilbof (approved):

issyl0 (approved):
Excellent, thank you!

PR #36, opened 2020-03-22
Add queston 10: are your basic needs being met

huwd:
## Screenshots
image

image


issyl0 (approved):
Looks good. According to the spreadsheet there's logic needed for when both "essential supplies" and "basic needs" are answered no, but we can do that later?

PR #38, opened 2020-03-22
Add Sentry

bilbof:
We need to add a SENTRY_DSN envvar for this to work.
We also need to create an app in the Sentry UI to get
the 'DSN'.

https://trello.com/c/dEL69YK8/67-set-up-sentry


issyl0:
Did you find someone with enough Sentry permissions to add a new project? I'm not that person, but it's probably worth asking again if it's been forgotten about?

bilbof:
@issyl0 not yet, I created another card for that bit https://trello.com/c/9gzHompQ/78-create-sentry-dsn

huwd (approved):

PR #41, opened 2020-03-22
Add error summary component for errors

NickColley:
Will need to test this more when we get more of the validation in, but for now this enables the error summary at the top of question pages to link to the field that has errored.

https://design-system.service.gov.uk/components/error-summary/


injms (approved):
I'm Ian James and I approve this pull request.

PR #42, opened 2020-03-22
Update i18n shape to nest questions in own key

injms:
To allow for easier plucking of _just_ the questions, all questions are now nested under the key `questions`.

No visual changes, changes to the way the app works - just to the way the internationalisation file is laid out.

The necessary views, controllers, helpers, and tests have been updated to use this new shape. Lots of repetitive changes mean that this is probably worth checking thoroughly.


bilbof (approved):

leenagupte (approved):
Amazing! Thanks! 🏅

PR #43, opened 2020-03-22
What is your nhs number

huwd:
# Screenshots
image

image

image

image

image

Also tested with my own NHS number (which i won't print here, it passes!)

### DIFF
--- a/app/controllers/coronavirus_form/nhs_number_controller.rb
+++ b/app/controllers/coronavirus_form/nhs_number_controller.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+class CoronavirusForm::NhsNumberController < ApplicationController
+  include ActionView::Helpers::SanitizeHelper
+  include FieldValidationHelper
+
+  def show
+    render "coronavirus_form/#{PAGE}"
+  end
+
+  def submit
+    nhs_number = sanitize(params[:nhs_number]).presence
+    session[:nhs_number] = nhs_number
+
+    invalid_fields = validate_mandatory_text_fields([:nhs_number], PAGE)
+
+    if invalid_fields.any?
+      flash.now[:validation] = invalid_fields
+      render "coronavirus_form/#{PAGE}"
+    elsif session["check_answers_seen"]
+      redirect_to controller: "coronavirus_form/check_answers", action: "show"
+    else
+      redirect_to controller: "coronavirus_form/#{NEXT_PAGE}", action: "show"
+    end
+
+  end
+private
+
+  PAGE = "nhs_number"
+  NEXT_PAGE = "check_answers" # TODO change to address
+
+  def previous_path
+    "/" # TODO
+  end
+end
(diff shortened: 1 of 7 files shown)


app/controllers/coronavirus_form/nhs_number_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  def clean_nhs_number(unclean_nhs_number)
+    unclean_nhs_number&.gsub(" ", "")
+  end
+
+  def validate_fields(nhs_number)
+    validation_response = validate_nhs_number_correctness(nhs_number)
+    return [] if validation_response[:valid]
+
+    puts validation_response.inspect
```
bevanloon: Should we remove this?

issyl0: Yeah, this is nowhere near ready yet, we'll tidy up later.

app/helpers/nhs_number_validator_helper.rb
```diff
@@ -0,0 +1,68 @@
+# frozen_string_literal: true
+
+module NhsNumberValidatorHelper
+  def validate_nhs_number_correctness(number)
+    puts number.inspect
+    if number.nil?
```
bevanloon: I know you've cleaned the number up above, but what about `.blank?` that will validate the empty string or a nil string or something like `"  "` if I remember rightly

huwd: ah, good shout, and we want the empty string to return too right?

app/controllers/coronavirus_form/nhs_number_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+        text: validation_response[:message],
+      },
+    ]
+  end
+
+  PAGE = "nhs_number"
+  NEXT_PAGE = "check_answers" # TODO change to address
+
+  def previous_path
+    "/" # TODO
```
issyl0:
```suggestion
    coronavirus_form_know_nhs_number_path
```

app/controllers/coronavirus_form/nhs_number_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+    [
+      {
+        field: "nhs_number",
+        text: validation_response[:message],
+      },
+    ]
+  end
+
+  PAGE = "nhs_number"
+  NEXT_PAGE = "check_answers" # TODO change to address
```
issyl0:
```suggestion
  NEXT_PAGE = "essential_supplies"
```

issyl0 (changes requested):
While we're here we may as well do the back/forwards routing.

app/controllers/coronavirus_form/nhs_number_controller.rb
issyl0: It's gone now.

issyl0 (approved):

config/locales/en.yml
```diff
@@ -178,9 +178,20 @@ en:
           label: Postcode
       basic_care_needs:
         title: Are your basic care needs being met at the moment?
-        hint: For example, washing and bathing yourself and keeping your house clean, with the help of friends or family.
+        hint: For example, washing and bathing yourself and keeping your house clean, with the help of friends or family.
```
```suggestion
        hint: For example, washing and bathing yourself and keeping your house clean, with the help of friends or family.
```

issyl0 (approved):
LGTM. 🚢

PR #54, opened 2020-03-22
Update routes and comments

huwd:
Rejiggers the routes to match the Sunday spreadsheet


issyl0 (approved):

issyl0:
Let's make these reflect teh current reality and then we can fix them up again in five minutes if there's a v5.

PR #72, opened 2020-03-22
Re order results page

huwd:
Also required removing disused yaml files..
... which then met removing other code and routes

# Screenshot
image


injms (approved):
This is indeed a pull request that should be merged in.

PR #76, opened 2020-03-22
Remove start page and redirect / to GOV.UK start page

issyl0:
- This matches the guidance in the Service Manual that every service should start at GOV.UK: https://www.gov.uk/service-manual/technology/get-a-domain-name#ensure-users-start-their-journey-on-govuk


bilbof (approved):
Does the translations file also need updating?

issyl0:
@bilbof I didn't realise we had a feature test for the start page. I completely removed it, I hope that's right. If you want me to test the redirect works then I can do, but my Cucumber is a bit rusty.

huwd (approved):

bilbof:
No worries, I added it mainly as a demo for future e2e tests.

issyl0:
> Does the translations file also need updating?

Good catch, thanks. That file shouldn't be any more complicated that it needs to be!

PR #79, opened 2020-03-22
Send users who don't live in England to ineligible page

leenagupte:
Trello: https://trello.com/c/AhNvbXVF/102-route-live-in-england-question-to-ineligible-page


huwd (commented):
Looks great,

Does the ineligible page have a back button?
If so, where does it take you - always back here?

leenagupte:
@huw that's a great question about the back button!

leenagupte:
@huw There's no back button on the prototype, and there isn't one here either.

huwd (approved):

PR #83, opened 2020-03-22
Update route for what is your nhs number

bevanloon:
Our check answers page assumes the route will be `nhs-number` owing to
the way the logic dasherises the "question" from the yaml file. Currently, it 404s from check answers page.

Ensure this is consistent in the routes.

Remove spurious whitespace in nhs number view.


huwd (approved):

huwd:
Thank you bevan! Great spot!

PR #84, opened 2020-03-22
Update the formatting to display answers better

injms:
Contact details, date of birth, and address all needed special consideration to make them more understandable.

## Desktop

image

## Mobile

image

### DIFF
--- a/app/controllers/coronavirus_form/check_answers_controller.rb
+++ b/app/controllers/coronavirus_form/check_answers_controller.rb
@@ -53,9 +53,9 @@ private
     joiner = " "
 
     if question.eql?("contact_details")
-      concatenated_answer << "Phone number: #{answer['phone_number_calls']}" unless answer["phone_number_calls"].nil?
-      concatenated_answer << "Text: #{answer['phone_number_texts']}" unless answer["phone_number_texts"].nil?
-      concatenated_answer << "Email: #{answer['email']}" unless answer["email"].nil?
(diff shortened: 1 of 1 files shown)


app/controllers/coronavirus_form/check_answers_controller.rb
```diff
@@ -36,20 +36,38 @@ def items
...
     end
   end
 
-  def concat_answer(answer)
-    answer.values.compact.join(" ")
+  def concat_answer(answer, question)
+    concatenated_answer = []
+    joiner = " "
+
+    if question.eql?("contact_details")
```
injms, line 55: I realise that the following is quite write-everything-twice, but I think it's a touch clearer for future people to see what we're doing here.

app/controllers/coronavirus_form/check_answers_controller.rb
```diff
@@ -36,20 +36,38 @@ def items
...
+      concatenated_answer << "Phone number: #{answer['phone_number_calls']}" unless answer["phone_number_calls"].nil?
```
leenagupte: You changed the `unless` and drop the `nil`
e.g.
```
concatenated_answer << "Phone number: #{answer['phone_number_calls']}" if answer["phone_number_calls"]
```

injms: Much better - thank you!

leenagupte (approved):
Looks good! There might be a neater way of doing this, but it's late in day, so go for it! 👍

PR #88, opened 2020-03-22
Change NHS number input type to number

### LINKED ISSUE #77: Usability: Should the 'NHS Number' text input use a numeric keyboard?

huwd:
Trello: https://trello.com/c/qLvXB4r5/106-change-nhs-number-input-to-type-numeric
Closes: https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/issues/77

## What
Make the input for NHS number numeric

Example rendered output:
```
<div class="govuk-form-group">  
  <label for="input-9468206c" class="gem-c-label govuk-label">
    NHS Number
  </label>
  <input name="nhs_number" class="gem-c-input govuk-input govuk-input--width-20" id="input-9468206c" type="text" inputmode="numeric" pattern="[0-9]*" value="">
</div>
```

## Why
Because our NHS colleages, tell us to wash our hands and use [numeric inputs for NHS numbers](https://service-manual.nhs.uk/design-system/patterns/ask-users-for-their-nhs-number).

Also [the design system](https://design-system.service.gov.uk/components/text-input/#numbers) and [publishing components](https://components.publishing.service.gov.uk/component-guide/input/numeric_input) make it easy.

### DIFF
--- a/app/views/coronavirus_form/nhs_number.html.erb
+++ b/app/views/coronavirus_form/nhs_number.html.erb
@@ -32,6 +32,7 @@
       label: {
         text: t('coronavirus_form.questions.nhs_number.nhs_number.label'),
       },
+      type: "number",
       error_message: error_items('nhs_number'),
       value: session.dig("nhs_number"),
       width: 20,


issyl0:
If this is the case do we need to strip out spaces? Does the `pattern` HTML let us input spaces? If people are on a computer they won't have the numeric keyboard pop up and they mgiht enter the spaces if they're reading a letter they have.

huwd:
Good thought, just checked inputting with spaces,
it allows it!

NickColley (changes requested):
> Do not use `<input type=”number”>` unless your user research shows that there’s a need for it. With `<input type=”number”>` there’s a risk of users accidentally incrementing a number when they’re trying to do something else - for example, scroll up or down the page. And if the user tries to enter something that’s not a number, there’s no explicit feedback about what they’re doing wrong.

We need to use:

` <input type="text" pattern="[0-9]*" inputmode="numeric" spellcheck="false">`

https://technology.blog.gov.uk/2020/02/24/why-the-gov-uk-design-system-team-changed-the-input-type-for-numbers/

Also needs testing on mobile iOS and Android

injms (approved):
💯

injms:
@nickcolley This change will give (id might be different):

```html
<input name="account-number" class="gem-c-input govuk-input" id="input-74b4a56c" type="text" inputmode="numeric" pattern="[0-9]*">
```

I think it's only missing the `spellcheck` attribute - would that be essential?

huwd:
@nickcolley i think that's what's output:
```
<input name="nhs_number" class="gem-c-input govuk-input govuk-input--width-20" id="input-9468206c" type="text" inputmode="numeric" pattern="[0-9]*" value="">
```

publishing components are [smart enough to take the numeric type](https://components.publishing.service.gov.uk/component-guide/input/numeric_input) and do the right thing by the design system.

I think it turns out that's also the answer to @issyl0's question

NickColley:
Ah okay, the component's API is misleading as we try to align with the DOM normally.

For spellcheck:
> If you are asking users for information which is not appropriate to spellcheck, like a reference number, name, email address or National Insurance number, disable the spellcheck.

injms:
Hmm, I we can't add the spellcheck attribute with the component. I'll add it as an issue to the Publishing Components repo, if it isn't already there - but I don't think we can add it here without hardcoding the input element.

I think the components API was left as is to preserve backwards compatibility.

huwd:
thanks both!

NickColley (approved):

PR #89, opened 2020-03-22
Always route users to enter NHS number

brucebolt:
Currently, if a user has seen the "check your answers" page and edits the response to whether they know their NHS number, they will be returned to the "check your answers" page.  This updates the redirect so users are forwarded to enter their NHS number, even if they have seen the "check your answers" page.

Trello card: https://trello.com/c/Q8aycv1O

### DIFF
--- a/app/controllers/coronavirus_form/know_nhs_number_controller.rb
+++ b/app/controllers/coronavirus_form/know_nhs_number_controller.rb
@@ -21,12 +21,12 @@ class CoronavirusForm::KnowNhsNumberController < ApplicationController
       flash.now[:validation] = invalid_fields
       log_validation_error(invalid_fields)
       render "coronavirus_form/#{PAGE}", status: :unprocessable_entity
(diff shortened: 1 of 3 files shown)


huwd (approved):

app/controllers/coronavirus_form/know_nhs_number_controller.rb
```diff
@@ -21,12 +21,12 @@ def submit
       flash.now[:validation] = invalid_fields
       log_validation_error(invalid_fields)
       render "coronavirus_form/#{PAGE}", status: :unprocessable_entity
+    elsif session[:know_nhs_number] == I18n.t("coronavirus_form.questions.know_nhs_number.options.option_yes.label")
```
leenagupte, line 24: I wonder if it would be easier to remove the `elsif session["check_answers_seen"]` condition, and just have the else route to `NEXT_PAGE`?

brucebolt: There is a scenario where a user edits their answers and selects 'no' (and has seen the 'check your answers' page), so I think we need to keep this condition there.

PR #95, opened 2020-03-22
Fix focus state on link on ineligible outside of England page

injms:
## What

Added `govuk-link` class to give link the correct focus state:

## Why

Link focus states should be consistent and accessible.

## Visual differences

Before:

Screenshot 2020-03-22 at 17 51 48

After:

Screenshot 2020-03-22 at 17 51 58


brucebolt (approved):

NickColley (approved):
good catch

PR #96, opened 2020-03-22
Ensure the NHS number doesnt get reformated

NickColley:
This happens on iOS and is a recommendation from the NHS team.

> add meta name="format-detection" content="telephone=no" to your page to stop iOS treating the NHS number as a phone number (but this means that iOS will not recognise any phone numbers on the page)

https://service-manual.nhs.uk/design-system/patterns/ask-users-for-their-nhs-number

### DIFF
--- a/app/views/coronavirus_form/know_nhs_number.html.erb
+++ b/app/views/coronavirus_form/know_nhs_number.html.erb
@@ -4,6 +4,8 @@
 
 <% content_for :meta_tags do %>
   <meta name="description" content="<%= t("coronavirus_form.questions.know_nhs_number.title") %>" />
+  <%# stop iOS treating the NHS number as a phone number %>
+  <meta name="format-detection" content="telephone=no" />
 <% end %>
 
 <% content_for :back_link do %>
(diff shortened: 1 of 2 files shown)


huwd (approved):

app/views/coronavirus_form/know_nhs_number.html.erb
```diff
@@ -4,6 +4,8 @@
 
 <% content_for :meta_tags do %>
   <meta name="description" content="<%= t("coronavirus_form.questions.know_nhs_number.title") %>" />
+  # stop iOS treating the NHS number as a phone number
```
brucebolt: These comments are being rendered to the user.  This will fix that.
```suggestion
  <%# stop iOS treating the NHS number as a phone number %>
```

brucebolt (changes requested):

app/views/coronavirus_form/know_nhs_number.html.erb
NickColley: Was too hasty with this one. Need to take a breather, thanks.

brucebolt (approved):
Change looks good 👍

NickColley:
Thanks, will slow down as well, thank you for your review.

PR #101, opened 2020-03-22
Remove /coronavirus-form from all routes

brucebolt:
Form questions, confirmation, etc will now be served from the root.


issyl0 (commented):
Ah you did it the easier way, whereas I tried to unknit all the controllers/views/routes/tests etc from `/coronavirus-form/...`!

leenagupte:
This is much nicer!

issyl0 (approved):

huwd (approved):

PR #103, opened 2020-03-22
Selectively load Sass

injms:
## What

Only importing the Sass that is needed by the app.

I've gone through and checked each question, plus the error state, the ineligible pages, plus the privacy notice and all look identical to me.

## Why

It'll reduce the amount of CSS users need to load, speeding up the service.

Before: 349.36KB of CSS
After: 116.87KB of CSS
With minification: 96.49KB of CSS

## Visual differences

None.


NickColley:
Worth testing each page and double checking but on paper this is a great change.

Gemfile
```diff
@@ -14,6 +14,7 @@ gem "log_formatter", "~> 0.8.2"
 gem "puma", "~> 4.3"
 gem "sentry-raven", "~> 3.0"
 gem "uglifier", "~> 4.2"
+gem "sassc-rails"
```
NickColley: think this'll need to be pinned for it to work

injms: argh, of course

injms: Now changed to `sass-rails` and pinned to less than 6.

app/assets/stylesheets/application.scss
```diff
@@ -1,4 +1,31 @@
-@import 'govuk_publishing_components/all_components';
+// @import 'govuk_publishing_components/all_components';
+
+@import 'govuk_publishing_components/govuk_frontend_support';
```
NickColley, line 3: inset text is missing

app/assets/stylesheets/application.scss
```diff
@@ -1,4 +1,31 @@
-@import 'govuk_publishing_components/all_components';
+// @import 'govuk_publishing_components/all_components';
+
```
NickColley, line 2: can you check you've got the styles needed for the confirmation page please?

NickColley:
Noticed two components required.

Would be good to get another person to visit every possible page in the journey and check nothing is missing before merging.

app/assets/stylesheets/application.scss
injms: thanks! added

injms:
> Noticed two components required.
> 
> Would be good to get another person to visit every possible page in the journey and check nothing is missing before merging.

I've added inset text - what was the other component?

NickColley:
Panel for the confirmation

NickColley:
@injms could you rebase your branch and double check all pages please :pray:

injms:
> @injms could you rebase your branch and double check all pages please 🙏

rebased and checked x2, including the final page and the ineligible pages.

Nooshu (approved):
Approve to be merged. I will check on staging once merged.

NickColley (approved):

PR #104, opened 2020-03-22
Increase puma workers/threads

bilbof:
This should increase the load each instance can handle.

I'm not worried about the amount of memory, since we have
2GB memory per instance.


huwd (approved):

huwd:
LGTM, I have a feeling @theseanything might know more about any consequences than me though?

theseanything (approved):

PR #107, opened 2020-03-22
Make hint text darker

### LINKED ISSUE #100: Ensure hint text is black
Tim suggestion since users of this service may be more likely to have visual impairments and we know this can be a problem sometimes.

NickColley:
The GOV.UK Design System team are aware of issues with lighter grey text for some disabled users
For example those with dyslexia or visual impairments.
Since this service is targetted at vulnerable people we're setting hint text to a darker colour to be safe.

This was in Tim Paul's original prototype and his recommendation.

Closes https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/issues/100

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -6,3 +6,10 @@
 .app-list--spaced > li:last-child {
     margin-bottom: govuk-spacing(0);
 }
+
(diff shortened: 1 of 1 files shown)


NickColley:
@injms if there's a way to do this by passing classes I would prefer that, but I could not see a way.

injms (approved):
Super - good call on making this change.

PR #113, opened 2020-03-22
Update en.yaml file with up-to-date text

m-green:
This updates the en.yaml file with the up-to-date content.

I may have made some yaml / html errors - thought it was best to raise the PR now rather than try to set up docker etc to preview locally. Hope that's ok!

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -244,7 +244,9 @@ en:
         title: Do you have any special dietary requirements?
         hint: |
           <p class="govuk-hint app-hint">For example, you're a vegetarian or you have a food allergy.</p>
-          <p class="govuk-inset-text">For now, food supplies will be delivered in standard packages. Check the ingredients lists carefully if you have a food allergy or follow a special diet. If you have special dietary requirements, some of the food may not meet them.</p>
+          <div class="govuk-inset-text">
+            <p class="govuk-body">For now, food supplies will be delivered in standard packages. Check the ingredients lists carefully if you have a food allergy or follow a special diet. If you have special dietary requirements, some of the food may not meet them.</p>
+          </div>
         options:
           option_yes:
             label: "Yes"
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -2,7 +2,7 @@ en:
   check_your_answers:
     title: Check your answers before sending your application
     heading: Now send your application
-    confirmation: By submitting this application you're confirming that, to the best of your knowledge, the details you're providing are correct.
+    confirmation: By submitting this application you’re confirming that, to the best of your knowledge, the details you’re providing are correct.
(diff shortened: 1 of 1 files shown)


issyl0 (commented):
General comment: We use typographical quotes (‘ instead of ') for apostrophes in a lot of places in this file but not everywhere, should we be consistent?

config/locales/en.yml
```diff
@@ -239,7 +242,10 @@ en:
         custom_select_error: Select yes if your basic care needs are being met at the moment
       dietary_requirements:
         title: Do you have any special dietary requirements?
-        hint: For example, you're a vegetarian or you have a food allergy.
+        hint: |
+          <p class="govuk-hint app-hint">For example, you're a vegetarian or you have a food allergy.</p>
```
nickcolley: The hints will now be black test so I would recommend changing `govuk-hint app-hint` to `govuk-body` and removing the <br>

config/locales/en.yml
```diff
@@ -239,7 +242,10 @@ en:
...
+          <br /><br />
+          <p class="govuk-inset-text">For now, food supplies will be delivered in standard packages. Check the ingredients lists carefully if you have a food allergy or follow a special diet. If you have special dietary requirements, some of the food may not meet them.</p>
```
NickColley, line 247:
```html
<div class="govuk-inset-text">
    <p class="govuk-body">...</p>
</div>
```

config/locales/en.yml
nickcolley: actually need to double check this.

config/locales/en.yml
m-green: Thanks Nick - is this correct now? https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/pull/113/commits/388cc68e88c2bc7bc4d0d44180cad19112000ab3

config/locales/en.yml
```diff
@@ -239,7 +242,11 @@ en:
         custom_select_error: Select yes if your basic care needs are being met at the moment
       dietary_requirements:
         title: Do you have any special dietary requirements?
-        hint: For example, you're a vegetarian or you have a food allergy.
+        hint: |
+          <p class="govuk-hint app-hint">For example, you're a vegetarian or you have a food allergy.</p>
+          <div class="govuk-inset-text">
```
NickColley: The component seems  wrap all this code with a div and govuk-body class which I did not expect.

So I think your previous revision was fine, sorry :man_facepalming:

NickColley (approved):

config/locales/en.yml
m-green: No worries - I've reverted it

issyl0 (approved):
This is looking great. Thanks Mark!

PR #114, opened 2020-03-22
Add postcode validation

### LINKED ISSUE #71: Validation: support address, check postcode regex isnt too clever
It looks a bit clever, we need to make sure it doesnt restrict input by making too many assumptions about postcodes.

brucebolt:
Adds validation for postcode entered on the support address question.  Maintains the existing conditional validation where a user can enter either a county or postcode or both.

Trello card: https://trello.com/c/EqVwf3E1

Fixes: #71


NickColley (approved):
If you're confident that regex is loose enough that it wont create barriers this looks good to me.

PR #117, opened 2020-03-22
Add visually hidden label to address form

### LINKED ISSUE #116: Support address second line does not have a label

NickColley:
This is what is recommended by the Design System.

Ensures that screen readers know what this field is for.

Closes https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/issues/116


brucebolt (approved):

injms (approved):
Nice catch - well done 👍

huwd (approved):
Thank you! Should have spotted that when i put it in!

PR #119, opened 2020-03-22
Make the asset pipeline apply cache control

tlwr:
What
---

Not sure if this is needed as I'm out of the loop on the CDN stuff:

- Add cache control headers to asset pipeline
- Cache assets for 24h
- This will get picked up by CloudFront, so we can keep DefaultTTL to 0

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -27,7 +27,15 @@ Rails.application.configure do
 
   # Disable serving static files from the `/public` folder by default since
   # Apache or NGINX already handles this.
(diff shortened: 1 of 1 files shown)


theseanything (approved):

NickColley:
Would be good to get Matt's approval if possible but we should get this merged before go live if he's not around.

Nooshu (approved):
Looks good to me @tlwr. Lighthouse will probably still complain as I think it expects 1 year. But realistically that won't be needed, and very few people will be needing to visit that many times. Plus browsers are aggressive in removing cached files, so they will never last the full time specified.

Issue #91, opened 2020-03-22
Review use of autocomplete and spellcheck attributes


NickColley:
Autocomplete can massively help disabled people and is a WCAG 2.1 requirement.

https://www.w3.org/TR/WCAG21/#input-purposes

The GOV.UK Design System examples include autocomplete / spellcheck attributes in their examples:

https://design-system.service.gov.uk/patterns

We should take care to only use it where we're confident autofill would be desirable.

NickColley:
Asked Tim P. There's some nuance as we can assume some carers may be submitting on behalf of someone else.

NickColley:
Added autocomplete, moving scope of this to spellcheck and demoting to 'enhancement'.

PR #123, opened 2020-03-22
Add autocomplete values where possible

NickColley:
Also includes setting the type for certain fields, which will improve
the keyboard usability on mobile.

Check the following patterns to confirm the changes are correct:

- [date of birth](https://design-system.service.gov.uk/patterns/dates/)
- [names](https://design-system.service.gov.uk/patterns/names/)
- [emails](https://design-system.service.gov.uk/patterns/email-addresses/)
- [phones](https://design-system.service.gov.uk/patterns/telephone-numbers/)
- [addresses](https://design-system.service.gov.uk/patterns/addresses/)

Improves https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/issues/91 , we cant use spellcheck=false as the input macros do not allow it yet.

### DIFF
--- a/app/views/coronavirus_form/name.html.erb
+++ b/app/views/coronavirus_form/name.html.erb
@@ -29,6 +29,7 @@
   name: "first_name",
   error_message: error_items("first_name"),
   value: session["name"]["first_name"],
+  autocomplete: "given-name",
 } %>
 
 <%= render "govuk_publishing_components/components/input", {
@@ -49,6 +50,7 @@
   name: "last_name",
   error_message: error_items("last_name"),
   value: session["name"]["last_name"],
+  autocomplete: "family-name",
 } %>
 <%= render "govuk_publishing_components/components/button", text: "Continue", margin_bottom: true %>
 <% end %>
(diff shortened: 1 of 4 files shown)


app/views/coronavirus_form/name.html.erb
```diff
@@ -29,6 +29,7 @@
   name: "first_name",
   error_message: error_items("first_name"),
   value: session["name"]["first_name"],
+  autocomplete: "given-name",
 } %>
 
 <%= render "govuk_publishing_components/components/input", {
```
injms, line 35: Should we add `autocomplete="additional-name"` for the middle name input?

injms (approved):
Not a blocker - should we add

- "bday-day"
- "bday-month"
- "bday-year"

to the date of birth page?

NickColley:
@injms does the date input not include these? Hmm made a wild assumption there

app/views/coronavirus_form/name.html.erb
injms: It's an optional field, so let's not.

injms (approved):

PR #125, opened 2020-03-22
Redirect to correctly if user doesn't know NHS number

leenagupte:
Trello: https://trello.com/c/FVbRtlJU/114-nhs-number-no-back-you-go-to-the-nhs-number-entry

Skip the "What is your NHS number" page, when the user clicks "back" if they've already
said that they don't know their NHS number.


issyl0 (approved):
Nice!

PR #127, opened 2020-03-22
Enhance date of birth validation

### LINKED ISSUE #44: Validation: date of birth, check if date is in the past
"Date of birth must be in the past"

Making a note to come back to this if I have time, not critical.

brucebolt:
Validates date of birth for the following:
- Negative numbers are not allowed
- Ensures date is not in the future
- Ensures all date fields are a number

Fixes #44

Trello card: https://trello.com/c/IJ9EXNlH

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -113,6 +113,7 @@ en:
       email_format: "Enter email address in the correct format, like name@example.com"
       postcode_format: "Enter a real postcode"
       negative_date: "Enter a real %{field}"
+      date_not_a_number: "Enter %{field} as a number"
     confirmation:
       title: "Registration complete"
       description: |
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -112,6 +112,7 @@ en:
       date_order: "Date of birth must be in the past"
       email_format: "Enter email address in the correct format, like name@example.com"
       postcode_format: "Enter a real postcode"
+      negative_date: "Enter a real %{field}"
     confirmation:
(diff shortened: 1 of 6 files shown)


config/locales/en.yml
```diff
@@ -110,6 +110,8 @@ en:
       date_order: "Date of birth must be in the past"
       email_format: "Enter email address in the correct format, like name@example.com"
       postcode_format: "Enter a real postcode"
+      negative_date: "Enter a real %{field}"
+      date_not_a_number: "Enter %{field} as a number"
```
NickColley, line 116: @m-green any opinions on this?

Example: "Enter day as a number"

spec/helpers/field_validation_helper_spec.rb
```diff
@@ -0,0 +1,74 @@
+require "spec_helper"
+
+RSpec.describe FieldValidationHelper, type: :helper do
+  context "#validate_date_of_birth" do
+    it "returns an error if year is blank" do
```
NickColley, line 5: Nice work on testing this

config/locales/en.yml
m-green: I think these came from a content designer's work on another form - so I think if they're happy, I'm happy.

brucebolt: This was all copied from the Brexit funding form.  I believe the logic was we couldn't use "Enter a real month", because someone entering "June" is still entering a real month albeit not the correct format.

bevanloon (approved):

PR #128, opened 2020-03-22
Set the nhs_number to nil if the user changes their answer

leenagupte:
Trello: https://trello.com/c/7nWLsSsg/121-remove-the-nhs-number-from-the-session-if-the-user-changes-their-answer-to-do-you-know-your-nhs-number

Remove the NHS number from the session if the answer to "Do you know your NHS number" is "No".

We were in a situation where the user initially answered "Yes" to "Do you know your NHS number", entered
their nhs number, and then on the check-answer screen, changed their answer "No", but the nhs number they'd
previously entered was still being displayed because it wasn't being cleared from the session.


brucebolt (approved):
👍

PR #126, opened 2020-03-22
Use back links on eligible pages

NickColley:
You may have made a mistake so replace the home link with a back link

Closes https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/issues/115

PR #129, opened 2020-03-22
Exclude all config from rubocop

leenagupte:
Should unblock: https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/pull/126

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -9,4 +9,5 @@ inherit_mode:
 
 Metrics/BlockLength:
   Exclude:
-    - config/routes.rb
+    - config/**/*
+    - config/environments/production.rb


.rubocop.yml
```diff
@@ -9,4 +9,5 @@ inherit_mode:
 
 Metrics/BlockLength:
   Exclude:
+    - config/environments/production.rb
```
issyl0, line 13: Can we shortcut and do `- config/**` instead?

issyl0 (commented):
Not blocking, but...

leenagupte:
@issyl0 updated.

huwd (approved):

NickColley (approved):

PR #131, opened 2020-03-22
Remove production config file from exclusions

leenagupte:
I had to add this in to get it passed the linter, but now it can be removed.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -10,4 +10,3 @@ inherit_mode:
 Metrics/BlockLength:
   Exclude:
     - config/**/*
-    - config/environments/production.rb


.rubocop.yml
```diff
@@ -9,5 +9,4 @@ inherit_mode:
 
 Metrics/BlockLength:
   Exclude:
-    - config/**/*
-    - config/environments/production.rb
+    - config/**/*s
```
NickColley: Is that cheeky 's' meant to be there?

leenagupte: Nooo! Good spot! Thanks.

NickColley (approved):

huwd (approved):

PR #134, opened 2020-03-22
Ensure error summary always links to first input

### LINKED ISSUE #132: Some error summarys linking to the form tag instead of the first input
In radio form groups, the field linked to is not the first input but instead the form itself.

It should focus the first input.

See http://govuk-frontend-review.herokuapp.com/full-page-examples/how-do-you-want-to-sign-in for an example of how the error summary should work.

NickColley:
The id was previously on the form tag itself, it needs to be on the
first input instead.

A few inputs such as the email and nhs number field needed hooking up
too.

This important to make sure the error summary works well with screen readers.

To test this change, run the application locally and check each validation message focuses (yellow outline) the right input.

Closes https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/issues/132


alex-ju (approved):
👍

PR #143, opened 2020-03-23
Show hint text everywhere it exists

issyl0:
Screenshot 2020-03-23 at 08 43 22
Screenshot 2020-03-23 at 08 43 39

### DIFF
--- a/app/views/coronavirus_form/name.html.erb
+++ b/app/views/coronavirus_form/name.html.erb
@@ -10,7 +10,14 @@
   <%= render "govuk_publishing_components/components/back_link", { href: previous_path } %>
 <% end %>
 
-<%= render "govuk_publishing_components/components/title", {
+<%= form_tag({},
+  "data-module": "track-coronavirus-form-vulnerable-people-name",
+  "data-question-key": "name",
+  "id": "name",
+  "novalidate": "true"
+) do %>
+
+  <%= render "govuk_publishing_components/components/title", {
   title: t("coronavirus_form.questions.name.title"),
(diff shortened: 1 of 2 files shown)


theseanything (approved):

issyl0:
I don't like the spacing between the title and the hint text on the name question. 🤔

app/views/coronavirus_form/name.html.erb
```diff
@@ -15,6 +15,10 @@
   margin_top: 0
 } %>
 
+<%= render "govuk_publishing_components/components/hint", {
```
leenagupte, line 25: Should this be inside the form_tag?

issyl0: Ah, I didn't copy 100% correctly from `app/views/contact_details.html.erb`. Yes, both the title and the hint are inside the form tag there. 😴One moment...

alex-ju:
> I don't like the spacing between the title and the hint text on the name question. 🤔

You can set a `margin_bottom: 3` on L16 if you want the spacing to be similar with the one on pages with radios.

timpaul:
> I don't like the spacing between the title and the hint text on the name question. 🤔

Agreed! Let's shrink the headings first and check again - we can apply spacing override classes if we need to

app/views/coronavirus_form/name.html.erb
issyl0: Done in b932ce2.

leenagupte (approved):

PR #145, opened 2020-03-23
Make check your answers more readable

NickColley:
Adjust CSS for the page to bring inline with prototype.

Makes sure that the key is longer so it can be read easier.

Turns off visited state for change links.


NickColley:
@alex-ju @injms SORRY for these overrides but I checked the view files and there is no way around this in such a short notice. I'll help upstream these TODOs when the heat goes down.

alex-ju (approved):
We'll try and address it properly after going live.

injms:
Curious to why we want to remove the links visited states?

NickColley:
They should always be visited in this user journey so they don't communicate anything useful.

But @timpaul may be able to give a better answer.

PR #146, opened 2020-03-23
Align question spacing across pages

alex-ju:
Align title/question spacing across pages.

Radios have a spacing of 3 units after the title/question. This brings all the other question pages in line with them.

Follow up on https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/pull/143#issuecomment-602461862


NickColley:
Need to check this change with Tim but I think it's a good idea.

issyl0 (approved):
Not a frontend dev, but I like this.

bilbof (approved):

huwd (approved):

timpaul:
> Need to check this change with Tim but I think it's a good idea.

Approved!

PR #148, opened 2020-03-23
Correctly replay NHS number

brucebolt:
This is being stored in the session hash using a symbol key, however we are attempting to replay it using a string key.  This does not always work in all cases.


issyl0 (approved):
Are there any other occurrences of this type of thing that we might have missed?

brucebolt:
@issyl0 Possibly, although nothing has been noticed in the user testing 🤔

PR #149, opened 2020-03-23
Strings to symbols

huwd:
Dependent on: Get :+1: from product before merging!

## What
Tracks down everywhere we'd made a key in the session object a `String`
and changes it to to a symbol

## Why
Switching between symbols and strings was the source of [this bug](https://stackoverflow.com/questions/8189416/why-use-symbols-as-hash-keys-in-ruby), after a [bit of chat on slack ](https://gds.slack.com/archives/C010JE8AMPZ/p1584955486216800?thread_ts=1584954108.200700&cid=C010JE8AMPZ), agreement that we should switch them all over as an after launch thing.

After a bit of reserach turns out symbols have some other benefits:
- [Symbols have access to ruby 2.0 keyword arguments](https://chriszetter.com/blog/2012/11/02/keyword-arguments-in-ruby-2-dot-0/)
- [Symbols are kind of like immutable strings and so are a tad faster](https://web.archive.org/web/20180709094450/http://www.reactive.io/tips/2009/01/11/the-difference-between-ruby-symbols-and-strings)

### DIFF
--- a/spec/controllers/coronavirus_form/name_controller_spec.rb
+++ b/spec/controllers/coronavirus_form/name_controller_spec.rb
@@ -34,7 +34,7 @@ RSpec.describe CoronavirusForm::NameController, type: :controller do
     end
 
 
-    it "sets session variables as sanatized symbolized keys" do
+    it "sets session variables as sanitize symbolized keys" do
       post :submit, params: params
       expect(session[session_key]).to eq person
     end
--- a/spec/controllers/coronavirus_form/name_controller_spec.rb
+++ b/spec/controllers/coronavirus_form/name_controller_spec.rb
@@ -47,9 +47,9 @@ RSpec.describe CoronavirusForm::NameController, type: :controller do
       }
 
       name = {
-        "first_name" => "Link",
-        "middle_name" => "Link",
-        "last_name" => "Link",
+        first_name: "Link",
+        middle_name: "Link",
+        last_name: "Link",
       }
 
       post :submit, params: params
--- a/spec/controllers/coronavirus_form/name_controller_spec.rb
+++ b/spec/controllers/coronavirus_form/name_controller_spec.rb
@@ -20,21 +20,21 @@ RSpec.describe CoronavirusForm::NameController, type: :controller do
   describe "POST submit" do
     let(:params) do
       {
-        "first_name" => "Harry",
-        "middle_name" => "Snape",
-        "last_name" => "Potter",
+        "first_name" => "Harry<script></script>",
+        "middle_name" => "Snape<script></script>",
+        "last_name" => "<script></script>Potter",
       }
     end
     let(:person) do
       {
-        "first_name" => "Harry",
-        "middle_name" => "Snape",
-        "last_name" => "Potter",
+        first_name: "Harry",
+        middle_name: "Snape",
+        last_name: "Potter",
       }
     end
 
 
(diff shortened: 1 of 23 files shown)


leenagupte:
@huwd I tried this out locally, and I got an error for fields that don't have a value.  So maybe use `try` anywhere there's a nested key? e.g. `session[:contact_details].try(:phone_number_calls)`

See: https://apidock.com/rails/Object/try
Screenshot 2020-03-23 at 11 51 43

spec/controllers/coronavirus_form/name_controller_spec.rb
```diff
@@ -16,21 +16,21 @@
...
-        "last_name" => "Potter",
+        first_name: "Harry",
+        middle_name: "Snape",
+        last_name: "Potter",
       }
     end
 
 
-    it "sets session variables" do
+    it "sets session variables as sanatized symbolized keys" do
```
leenagupte: Typo: should be "sanitize".

huwd:
Ah interesting we're not memoizing everywhere in the same way.
will amend that!

huwd:
@leenagupte how does .try work compared to .dig?

```ruby
x = {hello: {world: 1}}  
x[:hello][:world]
> 1
x[:hello].try(:world)
> nil
x[:hello].dig(:world)
> 1
x[:hello].dig(:non_existant_symbol)
> nil
```

spec/controllers/coronavirus_form/name_controller_spec.rb
huwd: ... thank you! corrected!

leenagupte (approved):
👍 💯 Looks good to me!

If you have time, it would be great if you could fix up some of your commits, but it's not a blocker.

bilbof (dismissed):
From my understanding, sessions [don't have indifferent access](https://www.ruby-forum.com/t/session-with-indifferent-access/54244), so it would be better to use strings wherever possible when interacting with `session`.

leenagupte:
> From my understanding, sessions [don't have indifferent access](https://www.ruby-forum.com/t/session-with-indifferent-access/54244), so it would be better to use strings wherever possible when interacting with `session`.

Is that because they're always stored as strings? https://stackoverflow.com/questions/10307412/should-i-use-quotes-or-colons-for-session-keys-in-ruby-on-rails

bilbof:
@leenagupte I have a feeling this is the behaviour, but I haven't tested it. I think you can store them using symbols or strings, but when you retrieve them later, the nested objects will be strings. Is that right?

```ruby
# first controller
session[:object] = { symbolised: :value }

# next controller
session[:object] == { "symbolised" => :value }
```

bilbof:
@leenagupte sorry, I was wrong there. Using symbols or strings seems to be fine. Maybe I was being inconsistent with the values I was calling them with, which this change would fix :)

```ruby
    session[:object] ||= { symbolised: :value }
    puts session[:object]
    puts session.dig(:object, :symbolised)
    puts session[:object][:symbolised]
```

leenagupte:
@huwd I've rebased this against master.

huwd:
> @huwd I've rebased this against master.

Legend!

brucebolt (approved):
This looks great :clap:.

I'm guessing we need to do the same for the business volunteering form?  :trollface:

PR #150, opened 2020-03-23
Compress JavaScript in production

NickColley:
Before: 644kb
After: 235kb

Audits reported a 77kb improvement after GZIP compression.

I tested this by enabling it in the development environment, is there a good way to test this change in a proper production environment?


bilbof (approved):
Nice one. Let's see how it looks in staging. You could also deploy it to Heroku beforehand.

NickColley:
GZIP stats

Staging: 55.5 KB
Live: 164 KB

So big improvement

PR #153, opened 2020-03-23
Update favicon

NickColley:
Curious issue, because when testing the previous PR I saw the favicon at http://localhost:5000/assets/favicon.ico :thinking:


injms (approved):
_The Curious Case of the Missing Favicon_ has been solved.

Nooshu:
Looks like there is already a `favicon.ico` in the [public folder](https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/tree/master/public). Although it is 0 bytes again.

PR #155, opened 2020-03-23
Return 200 for validation errors

bilbof:
These will get logged anyway if we have the log level set to info.

https://trello.com/c/tZRgrJPt/161-style-422-errors-nicely-since-were-serving-a-lot-of-them

### DIFF
--- a/spec/controllers/coronavirus_form/basic_care_needs_controller_spec.rb
+++ b/spec/controllers/coronavirus_form/basic_care_needs_controller_spec.rb
@@ -26,13 +26,13 @@ RSpec.describe CoronavirusForm::BasicCareNeedsController, type: :controller do
 
     it "validates any option is chosen" do
       post :submit, params: { basic_care_needs: "" }
-      expect(response).to have_http_status(:unprocessable_entity)
+
       expect(response).to render_template(current_template)
     end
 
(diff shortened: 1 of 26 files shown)


spec/controllers/coronavirus_form/basic_care_needs_controller_spec.rb
```diff
@@ -26,13 +26,13 @@
...
     it "validates any option is chosen" do
       post :submit, params: { basic_care_needs: "" }
-      expect(response).to have_http_status(:unprocessable_entity)
+
       expect(response).to render_template(current_template)
     end
 
     it "validates a valid option is chosen" do
       post :submit, params: { basic_care_needs: "<script></script>" }
-      expect(response).to have_http_status(:unprocessable_entity)
```
huwd, line 35: Does render templat imply a 200 or can we do `to_have_http_status(:success)` or something similar?

huwd (approved):

injms (changes requested):
I'd like to make sure that we are actually logging the validation errors before this is merged in and we lose the ability to tell what parts of the form that users are struggling with.

bilbof:
I'd prefer not to do this, since having 422s in the logs is quite useful. I'm going to close this for now.

PR #156, opened 2020-03-23
Redirect users to start of journey if they enter mid-way through

leenagupte:
Trello: https://trello.com/c/bmI1RJhe

# What's changed?
Adds a `before_action` to each controller to check that the user has at the least completed the first question.

The exemptions to this are the first question itself, and the privacy page, because users should be able to ready the privacy policy before proceeding.

# Why?
If a user comes in part way through, e.g. https://coronavirus-vulnerable-people.service.gov.uk/nhs-number, they answer that question then get redirected to the first question only at the "check your answers" page. We should redirect them to the first question earlier on so that they don't have to repeat part of their journey.

### DIFF
--- a/features/form.feature
+++ b/features/form.feature
@@ -11,9 +11,6 @@ Feature: Filling in the form
     And I click the "Continue" button
     Then I will be redirected to the "/nhs-letter" path
 
-  Scenario: Answers "Have you recently had a letter from the NHS about your situation as someone who’s extremely vulnerable to coronavirus?"
-    When I visit the "/nhs-letter" path
-    Then I can see a "Have you recently had a letter from the NHS about your situation as someone who’s extremely vulnerable to coronavirus?" heading
-    And I choose "Not sure"
-    And I click the "Continue" button
-    Then I will be redirected to the "/name" path
+  Scenario: Visits an intermediate question
+      When I visit the "/essential-supplies" path
+      Then I will be redirected to the "/live-in-england" path
--- a/spec/controllers/coronavirus_form/basic_care_needs_controller_spec.rb
+++ b/spec/controllers/coronavirus_form/basic_care_needs_controller_spec.rb
@@ -3,6 +3,8 @@
 require "spec_helper"
 
 RSpec.describe CoronavirusForm::BasicCareNeedsController, type: :controller do
+  include_examples "redirections"
+
   let(:current_template) { "coronavirus_form/basic_care_needs" }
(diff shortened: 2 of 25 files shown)


theseanything (dismissed):

spec/controllers/coronavirus_form/basic_care_needs_controller_spec.rb
```diff
@@ -3,11 +3,14 @@
 require "spec_helper"
 
 RSpec.describe CoronavirusForm::BasicCareNeedsController, type: :controller do
+  include_examples "redirections"
+
   let(:current_template) { "coronavirus_form/basic_care_needs" }
   let(:session_key) { :basic_care_needs }
 
   describe "GET show" do
```
huwd, line 11: Should all these get an additional test, something like:

```ruby
it "redirects to the first question if the first question is unanswered" do
  session[:live_in_england] = nil
  get :show
  expect(response).to redirect_to(live_in_england_path)
end
```

theseanything: Yup - so've moved them out in to a shared example and they are included via `include_examples "redirections"`

features/form.feature
```diff
@@ -3,10 +3,10 @@ Feature: Filling in the form
   As a user
   I want to be able to answer questions
 
-Scenario: Visits the basic care needs page
-    When I visit the basic care needs page
-    Then I can see the basic_care_needs page content
-    And I can see the basic_care_needs radio button options
+Scenario: Visits the live in England page
```
huwd: Do we have a feature test for the redirect?

Perhaps something like:
```ruby
Scenario: Lands half way through the form
  Given: I have not answered the live_in_england question
  When: I land on the NHS letter page
  Then: I am redirected to the live_in_england page
  And: I can see the live_in_england page content
```

huwd (changes requested):
Might need another couple of test cases?

spec/controllers/coronavirus_form/basic_care_needs_controller_spec.rb
huwd: Ah ha!

theseanything:
> Might need another couple of test cases?

Done added a feature test and a controller test for the privacy page.

huwd (approved):

issyl0:
Needs a rebase.

PR #161, opened 2020-03-24
Strip leading and trailing spaces from text fields

leenagupte:
Trello: https://trello.com/c/mAHsb30l

# Motivation
Sometimes leading and trailing spaces are added by autocompletes - which causes them to fail validation. The most common occurrence of this is on postcode fields.

See: https://twitter.com/puncakka1/status/1242391848150355969


richardTowers (approved):
Looks good to me - Travis seems to be 🙈'ing it though - not sure why.

PR #163, opened 2020-03-24
Smoke test that the CSS is still being served

richardTowers:
from the same path. See https://trello.com/c/cfGnjdZV/175-fix-issues-with-assets-404ing-during-rolling-deployments

We're only testing the application CSS as this is the only file which
would cause a major degradation in user experience. Other things
(javascript, images, fonts, print css) either change so rarely that
they're not going to be a problem, or gracefully degrade.


tlwr (approved):
Maybe set message to something in `smoke-test` task?

tlwr (approved):

PR #164, opened 2020-03-24
Strip leading and trailing spaces from text fields

richardTowers:
(re-raising this to try and get Travis to see it)

Trello: https://trello.com/c/mAHsb30l

Motivation
----------

Sometimes leading and trailing spaces are added by autocompletes - which causes
them to fail validation. The most common occurrence of this is on postcode
fields.

See: https://twitter.com/puncakka1/status/1242391848150355969


richardTowers:
Travis found the other one.

PR #165, opened 2020-03-24
Make it clearer to users that they still need to submit the application

timpaul:
Drop-out rates on this page are about 50% right now. We'd like to try adjusting the content on the page and see if that has any significant effect. We will need to know when this change has gone live, so we can accurately measure the effect on completion rates.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -3,7 +3,7 @@ en:
     label: Privacy
   check_your_answers:
     title: Are you ready to send your application?
-    hint: Check your answers first
+    hint: Check your answers first.
     heading: Now send your application
     confirmation: By submitting this application you’re confirming that, as far as you know, the details you’re providing are correct.
     submit: Accept and send
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -2,9 +2,10 @@ en:
   privacy_notice:
     label: Privacy
   check_your_answers:
-    title: Check your answers before sending your application
+    title: Are you ready to send your application?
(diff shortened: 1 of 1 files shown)


config/locales/en.yml
```diff
@@ -2,9 +2,10 @@ en:
   privacy_notice:
     label: Privacy
   check_your_answers:
-    title: Check your answers before sending your application
+    title: Are you ready to send your application?
+    hint: Check your answers first
```
timpaul:
```suggestion
    hint: Check your answers first.
```

36degrees: I think that this will need to be added to the view as well before it has any effect.

config/locales/en.yml
```diff
@@ -2,9 +2,10 @@ en:
   privacy_notice:
     label: Privacy
   check_your_answers:
-    title: Check your answers before sending your application
+    title: Are you ready to send your application?
+    hint: Check your answers first.
```
timpaul, line 6: This is intended to appear immediately below the H1. Are any changes required to other files to make this work?

config/locales/en.yml
timpaul: @leenagupte - this is beyond my knowledge :-/ Could someone else take a look please?

leenagupte: Looking now 👀

leenagupte: https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/pull/167

PR #171, opened 2020-03-25
Fix Concourse URL for this repo

benthorner:


issyl0 (approved):
Good catch, thanks Ben!

issyl0:
I skipped the tests as they also run on master and this was a README change. Travis will also be slower today as it's not a weekend. 😉

PR #172, opened 2020-03-25
Change Dynimoid logging level to :info

huwd:
Trello: https://trello.com/c/CpNKGtCb/184-ensure-sensitive-data-doesnt-go-to-logit-splunk

**DRAFT PENDING TEST ON STAGING**

# Test
- Get this on staging
- Noodle around
  - Normal browse behaviour
  - Cause an error
  - Submit a stand-in for PII
  - Cause validation error
- Logs should have No PII
- Logs should have validation errors
- Logs should have errors
- check the logs since deploy

# What
tweak the logging levels in production to prevent PII getting in, but ensure we still get useful logs on validation attempts

# Why
Keep cyber happy, keep debugging devs happy.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -62,7 +62,7 @@ Rails.application.configure do
 
   # Use the lowest log level to ensure availability of diagnostic information
   # when problems arise.
-  config.log_level = :debug
+  config.log_level = :info
 
   # Prepend all log lines with the following tags.
   config.log_tags = [:request_id]
--- a/config/initializers/dynamoid.rb
+++ b/config/initializers/dynamoid.rb
@@ -8,7 +8,6 @@ Dynamoid.configure do |config|
   config.access_key = ENV["AWS_ACCESS_KEY_ID"]
   config.secret_key = ENV["AWS_SECRET_ACCESS_KEY"]
   config.region = "eu-west-2"
-  config.logger.level = :error
 
   unless Rails.env.production?
     # [Optional]. If provided, it communicates with the DB listening at the endpoint - to communication with DynamoDB local.


theseanything:
Do turns out logs can have PII in them 😢 - We ended up having reduce the log level  https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/pull/141

theseanything (changes requested):
Turns out logs can have PII in them 😢 - We ended up having reduce the log level #141. I think Dynamoid is sharing the default Rails logger - so we are suppressing all Rails log with setting. Maybe need to make a separate logger? Think we need info level for rails as validation errors probably aren't "application errors".

richardTowers:
I suggest we pause the prod bit of the pipeline, merge this, let it deploy to staging, run through a couple of journeys, have a look through the logs for any PII, revert it if we're not happy, unpause the pipeline otherwise.

theseanything:
Hmm even at :info level I'm seeing logs such as:
```
Started POST "/contact-details" for ::1 at 2020-03-30 12:10:08 +0100
Processing by CoronavirusForm::ContactDetailsController#submit as HTML
  Parameters: {"authenticity_token"=>"Xq6KfiLO3qkUDHCmCv64JSO6RRSrdWFQfZXilSGlmqo9hMVMM4YjHKpYbHbDtVfrosCo7E/UbZ+g4m4AAeVGew==", "phone_number_calls"=>"07000000", "phone_number_texts"=>"07000000", "email"=>"user@email.com"}
```
Which means even if we set Dynamoid at a lower lever, we'd still log PII

theseanything:
But we might be able to use something like this: https://guides.rubyonrails.org/action_controller_overview.html#parameters-filtering

theseanything:
I've created another PR that might solve this issue https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/pull/205

PR #173, opened 2020-03-25
Run cucumber specs in the pipeline

richardTowers:
NOTE: There are some significant trade-offs that you need to consider
when reviewing this. I recommend a read of this description before looking at
the code.

We have some cucumber feature tests, which are a much nicer way of
testing that the application is working than `curl --fail`.

At the moment, these use cucumber-rails which handles starting a local
copy of the application, and provides nice Rails-y features like being
able to use path helpers (e.g. `live_in_england_path` instead of
`/live-in-england`).

You can't use cucumber-rails to test external websites, because it's
tightly hooked in to Rack. This means the tests weren't suitable for
running against something hosted in GOV.UK PaaS for example.

Instead, we can use cucumber and the apparition driver directly. This
lets us test external domains, but doesn't provide all the nice rails-y
bits like cucumber-rails does. That makes it painful to use locally
against a locally running app.

What I've done here is tried to get the best of both worlds:

If you run `rails cucumber` without setting `TEST_URL` you'll get
cucumber-rails, and things work like they did before.

If you run `TEST_URL=http://example.com rails cucumber` you'll get
cucumber with the apparition driver, and it will talk to an external
website.

The drawback of this is that the `step_definitions` will have to use
lowest-common-denominator features - if you try and use something from
cucumber-rails it will work locally, and in Travis but fail on
concourse.

I think that's probably a reasonable trade-off given how much better
these tests would be than curl, but happy to discuss.

Also note I've only tested a couple of pages - we should probably do the
rest soon.


issyl0 (approved):
This looks great. Very thorough and useful description too, thanks! RuboCop is moaning but apart from that 🚢.

bilbof (commented):
Awesome work 👍 🏅 

It might be nice to leave the rails cucumber tests in place, and run a smaller test suite as a smoke test using cucumber, which might give us the best of both worlds.

I don't think we'd need to run the whole feature test suite as the smoke test. Perhaps the cucumber smoke test could contain a few large e2e tests that run through the whole form, in order to catch integration issues (e.g. redis/db failures) and basic rendering errors, and have a larger set of feature tests that use cucumber-rails. What do you think?

features/support/env.rb
```diff
@@ -1,59 +1,15 @@
-# IMPORTANT: This file is generated by cucumber-rails - edit at your own peril.
```
richardTowers, line 1: > edit at your own peril.

I like to live dangerously 🕵

theseanything (approved):

theseanything:
Looks great! - I'd be happy to take the hit on rails magic ✨ for more deployment confidence 🚀

theseanything:
> I don't think we'd need to run the whole feature test suite as the smoke test. Perhaps the cucumber smoke test could contain a few large e2e tests that run through the whole form, in order to catch integration issues (e.g. redis/db failures) and basic rendering errors, and have a larger set of feature tests that use cucumber-rails. What do you think?

Think we can use feature flags to optionally select tests for smoke tests - but maybe something we can iterate on in another PR?

features/support/env.rb
```diff
@@ -1,59 +1,15 @@
...
-
-# Possible values are :truncation and :transaction
-# The :transaction strategy is faster, but might give you threading problems.
-# See https://github.com/cucumber/cucumber-rails/blob/master/features/choose_javascript_database_strategy.feature
-# Cucumber::Rails::Database.javascript_strategy = :truncation
+if ENV.key? "TEST_URL"
+  require "capybara/apparition"
+  require "capybara/cucumber"
+  require "capybara/rspec"
+  Capybara.default_driver = :apparition
```
theseanything, line 11: I think we might be able to add `Capybara.app_host = ENV["TEST_URL"]` here. This means we don't have to set a env var in development.

features/step_definitions/step_definitions.rb
```diff
@@ -0,0 +1,20 @@
+When "I visit the {string} path" do |path|
+  visit ENV.fetch("TEST_URL", "") + path
```
theseanything: WIth the suggested change to the config, I think this can become
```suggestion
  visit path
```

theseanything (changes requested):
Small tweak to help running this in development

theseanything (approved):

PR #174, opened 2020-03-25
Update the medical conditions question

m-green:
## What
Update the medical conditions question.

## Why
Our 2i reviewer suggested the change to make it clearer how to answer the question.


StephenGill:
Looks good to me from a content point of view 👍

theseanything (approved):

PR #177, opened 2020-03-25
Clarify medical conditions

m-green:
This is an update to the question about medical conditions, to clarify the second bullet on the list of medical conditions.


stevenjmesser (approved):
LGTM

PR #179, opened 2020-03-26
Improved feature tests

theseanything:
This adds feature tests for the following:
- Visiting the privacy policy
- Being ineligible
- Completing and submitting the form

### DIFF
--- a/features/form.feature
+++ b/features/form.feature
@@ -20,3 +20,29 @@ Feature: Filling in the form
     And I click the "Privacy" link
     Then I will be redirected to the "/privacy" path
     And I can see a "Privacy" heading
+
+  Scenario: Not eligible as lives outside England
+    When I visit the "/live-in-england" path
+    And I choose "No"
+    And I click the "Continue" button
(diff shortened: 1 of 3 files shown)


leenagupte (commented):
These tests look great! Thanks for adding them 🏅 🏆 

The only concern I have is that it will make it harder for content designers to update the I18n files, if we're relying on the same content in the tests.

Do you think it would be possible to compare against the I18n value instead? Something a bit like: https://thoughtbot.com/blog/better-tests-through-internationalization ?

theseanything:
> Do you think it would be possible to compare against the I18n value instead? Something a bit like: https://thoughtbot.com/blog/better-tests-through-internationalization ?

So I think the smoke tests won't have access to Rails magic i.e. the (cucumber-rails) gem, which makes it tricky to use the i18n values (i.e we don't have the t() methods available). But maybe we can have a look again in the future, if it causes us too much grief with content changes.

leenagupte (approved):

PR #181, opened 2020-03-26
Only include CSS that's required for error pages

### LINKED ISSUE #130: 404 and 500 error pages load entirety of GOV.UK Frontend / Design System CSS
The static error pages use a downloaded version of the GOV.UK Design System CSS - this should be replaced with only the styles needed for the page, which could be inlined.

injms:
## What 

The error pages were including all of the CSS from the GOV.UK Frontend design system. This has been changed to only include the styles needed for the pages. This reduced CSS from 87.4KB to 24.1KB.

The CSS filename has a hash of the file added to it to avoid a cache clash if it needed updating.

Fixes #130.

## Why

To make the (hopefully rarely seen) error pages faster to load.

## Visual differences

None.

## Local testing

Set `config.consider_all_requests_local` to `true` in `config/environments/development.rb` to see the errors when running the app locally.


public/404.html
```diff
@@ -8,7 +8,7 @@
   <meta http-equiv="X-UA-Compatible" content="IE=edge">
 
   <title>Page not found – GOV.UK</title>
-  <link rel="stylesheet" media="all" href="/govuk-frontend-3.6.0.min.css" />
```
alex-ju, line 11: XHTML ftw

alex-ju (approved):
Nice one, I'll use it in the other service too. Thank you for your investment!

theseanything:
Looks good to me! We're going to have to hold off merging until we fix our deployment pipeline :(

public/404.html
injms: You knows it.

injms:
> Looks good to me! We're going to have to hold off merging until we fix our deployment pipeline :(

Ah well - can you approve when it's okay to merge this in?

injms:
Unblocked because #208 and #219 have been merged in.

PR #183, opened 2020-03-26
Add more guards to the date of birth

leenagupte:
This might be overkill, but https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/pull/182 did not fix this error:
https://sentry.io/organizations/govuk/issues/1583162201/?environment=production&project=5170680

I'm not able to reproduce the problem locally, as the form doesn't let me submit an invalid or incomplete date. 
I'm not sure how this error is occurring, so I've added extra guards just in case.

### DIFF
--- a/app/helpers/answers_helper.rb
+++ b/app/helpers/answers_helper.rb
@@ -33,10 +33,16 @@ module AnswersHelper
     elsif question.eql?("support_address")
       answer.values.compact.join(",<br>")
     elsif question.eql?("date_of_birth")
-      Time.zone.local(answer["year"], answer["month"], answer["day"]).strftime("%d/%m/%Y") if answer.present?
+      Time.zone.local(answer["year"], answer["month"], answer["day"]).strftime("%d/%m/%Y") if complete_date?(answer)
     else
       answer.values.compact.join(" ")
     end
+  end
+
+  def complete_date?(date)
+    date.present? &&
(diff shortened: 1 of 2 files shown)


app/helpers/answers_helper.rb
```diff
@@ -33,12 +33,18 @@ def concat_answer(answer, question)
...
-      Time.zone.local(answer["year"], answer["month"], answer["day"]).strftime("%d/%m/%Y") if answer.present?
+      Time.zone.local(answer["year"], answer["month"], answer["day"]).strftime("%d/%m/%Y") if complete_date?(answer)
     else
       answer.values.compact.join(" ")
     end
   end
 
+  def complete_date?(date)
+    date.present? &&
+      %w(day month year).sort == date.keys.sort &&
```
theseanything:
```suggestion
      %w(day month year).each { |required_key| date.key?(required_key) }
```
Nit: might be clearer?

theseanything (approved):
Maybe some clarity around logic, but looks good!

app/helpers/answers_helper.rb
leenagupte: Nice!

leenagupte: So it turns out if you use `each` and the condition fails, it will return the list of keys, i.e. ["day", "month", "year"], but if you use `all?` instead it works as expected:

`%w(day month year).all? { |required_key| date.key?(required_key) }`

PR #188, opened 2020-03-27
Tell users there's no email or text confirmation

StephenGill:
Some users expect to get an email or text message, and wonder if the service has 'worked'.

This commit adds a line to the confirmation screen telling them that this won't happen.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -119,6 +119,7 @@ en:
     confirmation:
       title: "Registration complete"
       description: |
+        <p class="govuk-body">That's everything - you will not receive an email or text message to confirm your registration.</p>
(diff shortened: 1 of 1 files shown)


config/locales/en.yml
```diff
@@ -119,6 +119,7 @@ en:
     confirmation:
       title: "Registration complete"
       description: |
+        <p class="govuk-body">That's everything - you will not receive an email or text message to confirm your registration.</p>
```
```suggestion
        <p class="govuk-body">That’s everything - you will not receive an email or text message to confirm your registration.</p>
```

StephenGill: lol

config/locales/en.yml
```diff
@@ -119,6 +119,7 @@ en:
     confirmation:
       title: "Registration complete"
       description: |
+        <p class="govuk-body">That’s everything - you will not receive an email or text message to confirm your registration.</p>
```
richardTowers: There was some talk of preferring en dashes earlier.

```suggestion
        <p class="govuk-body">That’s everything – you will not receive an email or text message to confirm your registration.</p>
```

richardTowers (approved):

PR #189, opened 2020-03-27
Redirect to session expired page if a user's session expires

leenagupte:
Trello: https://trello.com/c/4vyXO6Ww

# What's changed?
Show a "session has expired" page when user's session expires.

# Why?
Users were being shown a generic rails "the change you wanted was rejected" page which was causing a lot of confusion.

# Expected changes
## Before
image

## After
Screenshot 2020-03-30 at 11 55 36

# Testing locally
change the `expire_after` value in the [session store config](https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/blob/master/config/initializers/session_store.rb) to `1.minute`

### DIFF
--- a/app/views/coronavirus_form/session_expired.html.erb
+++ b/app/views/coronavirus_form/session_expired.html.erb
@@ -1,0 +1,17 @@
+<% content_for :title do %><%= t('session_expired.title') %><% end %>
+<% content_for :meta_tags do %>
+  <meta name="description" content="<%= t('session_expired.title') %>" />
+<% end %>
+
+<%= render "govuk_publishing_components/components/title", {
+  title:t("session_expired.title"),
+  margin_top: 0
+} %>
+
+<%= sanitize(t('session_expired.body_text')) %>
+
+<%= render "govuk_publishing_components/components/button", {
+  text: "Start now",
+  start: true,
+  href: "/"
+} %>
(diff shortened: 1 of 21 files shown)


app/views/coronavirus_form/session_expired.html.erb
```diff
@@ -0,0 +1,17 @@
...
+
+<%= render "govuk_publishing_components/components/title", {
+  title:t("session_expired.title"),
+  margin_top: 0
+} %>
+
+<%= sanitize(t('session_expired.body_text')) %>
+
+<%= render "govuk_publishing_components/components/button", {
+  text: "Start now",
```
bilbof, line 14: Perhaps this text could be 'Restart form' or something similar?

bilbof (approved):
Looks perfect. Nice work 👍

app/views/coronavirus_form/session_expired.html.erb
leenagupte: Ah, I was given the content in the card: https://trello.com/c/4vyXO6Ww. I will ask though.

pixeltrix:
FYI, the 422 errors you're seeing relate to this issue: https://github.com/rails/rails/issues/21948

leenagupte:
> FYI, the 422 errors you're seeing relate to this issue: [rails/rails#21948](https://github.com/rails/rails/issues/21948)

Yeah, I know, but I wasn't sure how else to detect the problem and be able to redirect users to a help page.

theseanything (approved):

PR #190, opened 2020-03-28
Only run 1 smoke test against prod at any one time

jpluscplusm:
This PR makes sure multiple `smoke-test-prod` jobs stack up, and don't run all at once.

Whilst this is proposed in order to enable a later PR to trigger the smoke test periodically, with timeouts, this PR can safely be merged (after review!) as the current pipeline behaviour has serial-ness implied by this job's dependency on `deploy-to-prod` which is itself `serial: true`


richardTowers (approved):
Looks good to me - the smoke test is very fast compared to the deploy, so I can't see how this could make anything worse.

Thanks @jpluscplusm!

PR #191, opened 2020-03-28
Make prod smoke tests fail if they take too long

jpluscplusm:
Production smoke tests can currently run forever without failing.

This PR places an arbitrary (but hopefully way-too-long) 5 minute timeout on the prod smoke test job.

It kinda-sorta builds on #190, but could be merged seperately.


szieglerICF:
The idea behind the change makes sense and the code looks good. Why does it run "forever" sometimes?

richardTowers (approved):
Given what the task does (`curl --fail`) 5m should be more than enough time. Seems like a good first step.

PR #194, opened 2020-03-28
Make smoke tests follow HTTP redirects

jpluscplusm:
Currently, at least 1 of the 2 invocations of the smoke test script (prod smoke test CD job) are testing a URL which emits highly-CDN-cacheable redirects to the public-facing gov.uk URL.

Without this`--location` it's possible the redirects would be correctly served from cache, but the app itself could be dead, but this script would not notice.

This PR makes the smoke test follow all 3XX responses and assert the 2XX-ness of the URL being consumed by the public.

### DIFF
--- a/concourse/tasks/smoke-test.yml
+++ b/concourse/tasks/smoke-test.yml
@@ -16,4 +16,4 @@ run:
       echo "$MESSAGE"
 
       # TODO we can come up with a more thorough test than this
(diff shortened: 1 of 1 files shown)


richardTowers:
Ah, good spot. The "tests" should probably be looking at a route that doesn't redirect to be honest. I'll raise a PR to point them at `/live-in-england`, which is the first route in the journey.

richardTowers:
Superseded by #202

PR #197, opened 2020-03-28
Use mini_racer in preference to therubyracer

pixeltrix:
The gem 'therubyracer' is dependent on an old version of libv8 with multiple security vulnerabilities and doesn't build easily on the latest version of macOS. Rails replaced it with the 'mini_racer' gem in the PR rails/rails#29285.

Feel free to ignore this PR if you're too busy - just looking to see if I could help and hit build problems with the old libv8 on macOS Catalina.


richardTowers (approved):
Thanks! Looks good to me.

If there's nobody brave enough to merge it over the weekend we'll get to it on Monday 🚢

issyl0:
Thanks again, @pixeltrix! Our work laptops still run Mojave so we're a bit behind the curve!

PR #198, opened 2020-03-28
Container image to run feature tests in pipeline

richardTowers:
Build a container that includes:

- ruby
- google chrome
- the application source code
- the application's dependencies

This is everything we need to run the feature tests against a remote URL
(previously we couldn't do this because we didn't have a container with
both ruby and chrome in).

We can get concourse to build this and upload it to the private Amazon
ECR (elastic container registry) that comes with the govuk-tools
concourse team (this is specific to GDS RE's concourse - see
https://reliability-engineering.cloudapps.digital/continuous-deployment.html#services)

Once we've got an image, we can use that to run the feature tests
against the staging URL.

We shouldn't run these tests against production yet, because they submit
the form, so they'd end up polluting live data. We could flag them so
they don't submit in staging though.

It's possible that building this image on every commit to master (and
blocking the pipeline on a clean build of the image) might turn out to
slow things down unacceptably. If that's the case we can look at caching
a base image or some other optimisation.

### DIFF
--- a/concourse/pipeline.yml
+++ b/concourse/pipeline.yml
@@ -28,7 +28,24 @@ resources:
       branch: master
       pro: true
 
+  - name: feature-tests-image
+    type: docker-image
+    icon: docker
+    source:
+      repository: ((readonly_private_ecr_repo_url))
+      tag: govuk-coronavirus-vulnerable-people-form-feature-tests
+
 jobs:
+  - name: build-feature-tests-image
+    serial: true
+    plan:
+      - get: govuk-coronavirus-vulnerable-people-form
+        trigger: true
+      - put: feature-tests-image
+        params:
+          build: govuk-coronavirus-vulnerable-people-form
+          dockerfile: govuk-coronavirus-vulnerable-people-form/features/Dockerfile
+
   - name: deploy-to-staging
     serial: true
     plan:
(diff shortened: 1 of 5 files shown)


concourse/pipeline.yml
```diff
@@ -28,7 +28,24 @@ resources:
...
+    source:
+      repository: ((readonly_private_ecr_repo_url))
+      tag: govuk-coronavirus-vulnerable-people-form-feature-tests
+
 jobs:
+  - name: build-feature-tests-image
+    serial: true
+    plan:
+      - get: govuk-coronavirus-vulnerable-people-form
+        trigger: true
```
issyl0, line 61: Building this on every commit feels excessive. It's not going to change that much. Can we put it on a schedule instead?

richardTowers: We could have a second git resource for just the bits of the repo that might change the image (I think `Gemfile*`, `.ruby-version` and `features/*`), or yeah, we could just do it daily.

richardTowers: Had a stab at that in f50d9a3 - untested though. Could either merge this and deal with any mistakes afterwards or point the pipeline at this branch (with prod paused). Whaddyathink?

issyl0: Merge and deal with mistakes afterwards is easier I reckon.

issyl0 (approved):

PR #184, opened 2020-03-27
Use path helpers in the redirects

leenagupte:

PR #199, opened 2020-03-28
Remove the PAGE and NEXT_PAGE constants

pixeltrix:
The PAGE constant can be replaced by either the controller_name for the field name or controller_path for the template name. The NEXT_PAGE constant can be replaced by named url helpers which are faster to generate than options passed to url_for (which is what happens internally inside the redirect_to method).

This conflicts with #184 which uses path helpers for the redirect_to but url helpers are preferred because the first check inside redirect_to is for a string that starts with a protocol and then it next checks if the argument is a string and if so prepends the protocol and host name with the port. This technically makes it slightly faster but you'd never notice in reality.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -6,7 +6,9 @@ class ApplicationController < ActionController::Base
 
   before_action :check_first_question, only: [:show]
 
-  def show; end
+  def show
+    render controller_path
+  end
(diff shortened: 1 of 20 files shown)


leenagupte (approved):
This is great! Thanks for helping us to clean all this up. Much appreciated! 🥇

PR #203, opened 2020-03-30
Add accessibility statement

injms:
## What

Add accessibility statement content and link to statement in the footer. Content from the ticket: https://trello.com/c/xnRaNRXL

Also removed label / title duplication in translation file for the footer links. The privacy policy has a label in one part of the translation file, and the title in another part of the file - both of which were the same.

## Why

The service needs a statement. 

## Visual changes

image


leenagupte (commented):
Looks great! All that's missing is a basic test. Something like: https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/blob/master/spec/controllers/coronavirus_form/privacy_controller_spec.rb

richardTowers (approved):

leenagupte (approved):
💯

PR #205, opened 2020-03-30
Improve logging of validation errors and requests

theseanything:
closes: #172 
Trello: https://trello.com/c/CpNKGtCb/71-ensure-sensitive-data-doesnt-go-to-logit-splunk

We currently have a logging level set to error, as we wanted to prevent accidentally logging PII information (which can be seen in request and Dynamoid logs). However, this has prevented us from capturing information about which validation errors users are experience as they are logged at an info level. This PR increases the logging level of the default Rails logger to info, but suppresses PII data by:
- Dynamoid uses a separate logger at a lower level (error)
- Parameter filter to mask user data with in requests.

### DIFF
--- a/config/initializers/dynamoid.rb
+++ b/config/initializers/dynamoid.rb
@@ -8,6 +8,7 @@ Dynamoid.configure do |config|
   config.access_key = ENV["AWS_ACCESS_KEY_ID"]
   config.secret_key = ENV["AWS_SECRET_ACCESS_KEY"]
   config.region = "eu-west-2"
+  config.logger = ActiveSupport::Logger.new(STDOUT)
   config.logger.level = :error
 
   unless Rails.env.production?
--- a/config/initializers/filter_parameter_logging.rb
+++ b/config/initializers/filter_parameter_logging.rb
@@ -3,4 +3,4 @@
 # Be sure to restart your server when you modify this file.
 
 # Configure sensitive parameters which will be filtered from the log file.
-Rails.application.config.filter_parameters += [:password]
+Rails.application.config.filter_parameters += %i[password email name first_name middle_name last_name nhs_number nhs_letter date_of_birth carry_supplies live_in_england phone_number_calls phone_number_texts know_nhs_number basic_care_needs essential_supplies medical_conditions dietary_requirements building_and_street_line_1 building_and_street_line_2 town_city county postcode]
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -62,7 +62,7 @@ Rails.application.configure do
 
   # Use the lowest log level to ensure availability of diagnostic information
   # when problems arise.
-  config.log_level = :debug
+  config.log_level = :info
 
   # Prepend all log lines with the following tags.
   config.log_tags = [:request_id]


config/initializers/filter_parameter_logging.rb
```diff
@@ -3,4 +3,4 @@
 # Be sure to restart your server when you modify this file.
 
 # Configure sensitive parameters which will be filtered from the log file.
-Rails.application.config.filter_parameters += [:password]
+Rails.application.config.filter_parameters += %i[password email name first_name middle_name last_name nhs_number nhs_letter date_of_birth carry_supplies live_in_england phone_number_calls phone_number_texts know_nhs_number basic_care_needs essential_supplies medical_conditions dietary_requirements building_and_street_line_1 building_and_street_line_2 town_city county postcode]
```
leenagupte, line 6: Is this supposed to filter out everything or just personally identifiable information? For example, do the boolean values need to be filtered out?

config/initializers/dynamoid.rb
```diff
@@ -8,6 +8,7 @@
   config.access_key = ENV["AWS_ACCESS_KEY_ID"]
   config.secret_key = ENV["AWS_SECRET_ACCESS_KEY"]
   config.region = "eu-west-2"
+  config.logger = ActiveSupport::Logger.new("log/dynamoid-#{Rails.env}.log")
```
theseanything: This needs to write to STDOUT

config/initializers/filter_parameter_logging.rb
theseanything: Hmm, probably not, but I didn't see I reason to leave it in. I was leaning towards "better safe than sorry".  But happy to remove some of them if you think it'd help as extra information in the logs.

huwd:
@theseanything does this replace this one?
https://github.com/alphagov/govuk-coronavirus-business-volunteer-form/pull/121

Would be interested to know if this is also the solution to the bis form stuff?

config/initializers/filter_parameter_logging.rb
leenagupte: To be honest I'm not sure if having the information would help or not.
The only concern I really have is with keeping this list up to date.

theseanything:
> @theseanything does this replace this one?
> [alphagov/govuk-coronavirus-business-volunteer-form#121](https://github.com/alphagov/govuk-coronavirus-business-volunteer-form/pull/121)
> 
> Would be interested to know if this is also the solution to the bis form stuff?

Maybe adding the filtered parameters, but you shouldn't have to make a separate logger as the biz form is no longer using Dynamoid.

config/initializers/filter_parameter_logging.rb
theseanything: yeah - had this concern too. Was thinking if there was a way to not log parameters at all, or how to add test to catch new parameters.

richardTowers: Yeah, I agree - this feels like it could lead to a dangerous game of whackamole where we add new questions and forget to ignore them.

huwd: does ruby code in initalizers have access to code in controller modules?

No chance we could do something like pull out the check your answers page and map out all the possible options?

That way we'd never to whack-a-mole?

pixeltrix: One possible option is to use the [lograge gem][1] - that will remove the params from the log line by default and you can customise the config to add in information that you need

[1]: https://github.com/roidrage/lograge

theseanything: Yeah! I looked at adding lograge - https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/pull/214

theseanything:
Closing in favour of #214

PR #208, opened 2020-03-30
Use asset sync to store compiled assets in S3

theseanything:
When we do a new release, we replace one instance at a time until we've replaced all the instances.

This means there's an old and a new version of the app running at the same time.

The new version of the app may decide that the assets are served from a different filename so it will render `<link>` tags with that filename. When the user's browser requests that URL, it might get load balanced to an application that doesn't know about the new file yet, resulting in a 404.

To overcome this we can store the assets in a S3 bucket and use that as our cloud front origin. This PR add the `asset-sync` gem to handle uploading assets to S3 and set the require environment variable in concourse.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,8 +6,10 @@ source "https://rubygems.org"
 
 gem "rails", "~> 6.0.2"
 
+gem "asset_sync"
 gem "bootsnap", "~> 1"
 gem "dynamoid"
+gem "fog-aws"
 gem "govuk_app_config", "~> 2.1.1"
 gem "govuk_publishing_components", "~> 21.31.0"
 gem "log_formatter", "~> 0.8.2"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -61,6 +61,11 @@ GEM
     apparition (0.5.0)
       capybara (~> 3.13, < 4)
       websocket-driver (>= 0.6.5)
+    asset_sync (2.11.0)
+      activemodel (>= 4.1.0)
+      fog-core
+      mime-types (>= 2.99)
+      unf
     ast (2.4.0)
     awesome_print (1.8.0)
     aws-eventstream (1.0.3)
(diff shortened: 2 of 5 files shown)


richardTowers (approved):
Looks like a good first pass to me! This won't actually serve the assets from S3 - there's a further change we need to make to CloudFront for that to happen. We can deploy this and check that the right assets are in place though.

I guess this will attempt to upload the files every time an instance starts, which given that we're running 30 instances is not ideal, but should be fine as a starting point. If we don't notice it causing any problems / slow downs it might be fine for the long term.

theseanything:
> I guess this will attempt to upload the files every time an instance starts, which given that we're running 30 instances is not ideal, but should be fine as a starting point. If we don't notice it causing any problems / slow downs it might be fine for the long term.

Ah sorry I forgot to describe when it runs! Turns out asset-sync hooks in with the `assets:precompiled` rake task, so it will run during staging (when the buildpack runs). So it should only be once per deploy.

richardTowers:
Ah nice, I was confused by the code being in `config/initializers` but I guess these run in the rake tasks too?

theseanything:
> Ah nice, I was confused by the code being in `config/initializers` but I guess these run in the rake tasks too?

So _I think_ the rails application is initialized when you run a rake task i.e. not exclusive to the `rails server` command

pixeltrix:
> So I think the rails application is initialized when you run a rake task i.e. not exclusive to the rails server command

Just to confirm for @theseanything's peace of mind - only rake tasks that have the `:environment` task as a dependency will initialize the Rails application, e.g.

``` ruby
task mytask: :environment do
  ## application initialized
end
```

vs.

``` ruby
task :mytask do
  ## application not initialized
end
```

You can see the asset_sync gem does the former here:
https://github.com/AssetSync/asset_sync/blob/master/lib/tasks/asset_sync.rake#L4

What the `:environment` task does is to load the `config/environment.rb` file which calls `Rails.application.initialize!`, which will obviously then load the initializers.

theseanything:
@pixeltrix Ah awesome, thanks for the explanation!

PR #214, opened 2020-03-31
Improve logging using lograge

theseanything:
closes: #172 
Trello: https://trello.com/c/CpNKGtCb/71-ensure-sensitive-data-doesnt-go-to-logit-splunk

We currently have a logging level set to error, as we wanted to prevent accidentally logging PII information (which can be seen in request and Dynamoid logs). However, this has prevented us from capturing information about which validation errors users are experience as they are logged at an info level. This PR increases the logging level of the default Rails logger to info, but suppresses PII data by:
- Dynamoid uses a separate logger at a lower level (error)
- Used the lograge gem to format logs and leaves out parameter that contain sensitive data

Lograge also has the added benefit of formatting the logs in JSON which maybe be easier to parse in our log collection systems.

logs look like this:
```
{"method":"GET","path":"/live-in-england","format":"html","controller":"CoronavirusForm::LiveInEnglandController","action":"show","status":200,"duration":1230.82,"view":1230.04,"time":"2020-03-31 13:48:45 UTC"}
{"method":"GET","path":"/live-in-england","format":"html","controller":"CoronavirusForm::LiveInEnglandController","action":"show","status":200,"duration":815.58,"view":815.19,"time":"2020-03-31 13:48:45 UTC"}
{"method":"POST","path":"/live-in-england","format":"html","controller":"CoronavirusForm::LiveInEnglandController","action":"submit","status":302,"duration":1.5,"view":0.0,"location":"http://localhost:5000/nhs-letter","time":"2020-03-31 13
:49:01 UTC"}
{"method":"GET","path":"/nhs-letter","format":"html","controller":"CoronavirusForm::NhsLetterController","action":"show","status":200,"duration":86.67,"view":86.09,"time":"2020-03-31 13:49:01 UTC"}
validation error - Select if you received the letter from the NHS
```


Gemfile
```diff
@@ -11,6 +11,7 @@ gem "dynamoid"
 gem "govuk_app_config", "~> 2.1.1"
 gem "govuk_publishing_components", "~> 21.31.0"
 gem "log_formatter", "~> 0.8.2"
```
richardTowers: Might be able to remove log_formatter and use lograge's JSON thing? I tried using lograge originally, but I couldn't get it to work - I ended up raging and using this other thing.

pixeltrix: This worked for me locally after removing `log_formatter` from the Gemfile:

```
config.lograge.enabled = true
config.lograge.formatter = Lograge::Formatters::Json.new
```

This gives the following log line:

```
{"method":"GET","path":"/live-in-england","format":"html","controller":"CoronavirusForm::LiveInEnglandController","action":"show","status":200,"duration":484.42,"view":483.82}
```

I thought you might need to ignore `/healthcheck` but looks like lograge already ignores that because it doesn't find a controller.

richardTowers: 🤦‍♂ I think I might have forgotten `config.lograge.enabled = true` last time I tried this. I was tired.

theseanything: Ah I guess I'm a little confused what output we need - do we need JSON? it didn't appear to  logging in JSON before 🤔 (or maybe it was)

theseanything: But I'll go ahead and remove this formatter anyway

pixeltrix: > Ah I guess I'm a little confused what output we need - do we need JSON? it didn't appear to logging in JSON before 🤔 (or maybe it was)

I think currently it's not logging anything but errors due to the setting here:
https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/blob/2eef96ff556d685b649919baceff20c50233b0a5/config/initializers/dynamoid.rb#L11

That's mutating the level on the default `Rails.logger`.

If the logs are going into Splunk then the default key/value pairs is the best option:
https://dev.splunk.com/enterprise/docs/developapps/logging/loggingbestpractices/#Use-clear-key-value-pairs

richardTowers: The logs are being shipped to Splunk and [Elasticsearch / Logstash / Kibana], so JSON would be nice (otherwise we'll have to change our Logstash filters I think). It was using `log_formatter/ruby_json_formatter` before, but that was only because I couldn't get lograge to do the right thing (and that was probably just me being stupid).

We might also want to consider not shipping to ELK though - not sure how useful it is given we built most of our dashboards in splunk.

pixeltrix: There's an explicit Logstash formatter for lograge if that's what's needed:
https://github.com/roidrage/lograge/blob/master/lib/lograge/formatters/logstash.rb

(note the requirement for the logstash-event gem)

config/initializers/dynamoid.rb
```diff
@@ -8,6 +8,7 @@
   config.access_key = ENV["AWS_ACCESS_KEY_ID"]
   config.secret_key = ENV["AWS_SECRET_ACCESS_KEY"]
   config.region = "eu-west-2"
+  config.logger = ActiveSupport::Logger.new(STDOUT)
   config.logger.level = :error
```
huwd, line 12: This might not be necessary
```suggestion
```

huwd (approved):

config/initializers/dynamoid.rb
huwd: actually, nah

richardTowers (approved):

PR #215, opened 2020-03-31
Move medical conditions question to earlier in the flow

brucebolt:
This moves the medical condition eligibility question to earlier in the flow, as follows:
1. Do you live in England?
2. Have you recently had a letter from the NHS?
3. Do you have a qualifying medical condition?
4. What is your name?
5. Form flow continues as previously

Trello card: https://trello.com/c/QT7Jr2qL

### DIFF
--- a/app/controllers/coronavirus_form/know_nhs_number_controller.rb
+++ b/app/controllers/coronavirus_form/know_nhs_number_controller.rb
@@ -31,6 +31,6 @@ class CoronavirusForm::KnowNhsNumberController < ApplicationController
 private
 
   def previous_path
-    medical_conditions_path
+    contact_details_path
   end
 end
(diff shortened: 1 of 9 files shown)


app/controllers/coronavirus_form/know_nhs_number_controller.rb
```diff
@@ -31,6 +31,6 @@ def submit
 private
 
   def previous_path
-    medical_conditions_path
+    name_path
```
leenagupte: I think this should be: `contact_details_path`

leenagupte (commented):
I think only one back link is going to the wrong place, other than that, it looks good.

app/controllers/coronavirus_form/know_nhs_number_controller.rb
brucebolt: Good spot.  We should probably have a feature test for the back links.  It has caught us out (on the business volunteering form) a few times.

leenagupte (approved):
👍 💯

PR #216, opened 2020-03-31
Restringify some keys

leenagupte:
Follows on from: https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/pull/149

Rails unhelpfully stringifies some keys in the session:
https://github.com/rails/rails/blob/ea4f0e2baba8653b03fba154357842933cf7b778/actionpack/lib/action_dispatch/middleware/session/cookie_store.rb#L97

Which means that our views weren't able to access the values.

This only seems to affect nested keys.

### DIFF
--- a/app/helpers/answers_helper.rb
+++ b/app/helpers/answers_helper.rb
@@ -26,14 +26,14 @@ module AnswersHelper
 
     if question.eql?("contact_details")
       concatenated_answer = []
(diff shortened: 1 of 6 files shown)


issyl0:
These need to be double quotes for RuboCop to be happy.

issyl0:
Oh wait, they are, the singles are already in double quotes for strings so that's fine. This is why RuboCop is a robot not a human. I can't read.

issyl0 (approved):

PR #218, opened 2020-04-01
Fix typo in set_pipeline config

richardTowers:
Discovered this when I tried to set-pipeline the pipeline.


issyl0 (approved):
Damn. I did look at the diff, but that obviously doesn't do any validation.

PR #223, opened 2020-04-02
Revise routes annotations

huwd:
A part of: [If a user goes back and chooses "no" to a question they previously chose "yes" to, remove the other data held against that question](https://trello.com/c/AnGnIZ7I/15-if-a-user-goes-back-and-chooses-no-to-a-question-they-previously-chose-yes-to-remove-the-other-data-held-against-that-question)

## What

Revises the annotations for the routes file, amends one where the question has been updated, and removes some versioning information we put up at a time 

## Why

I and other devs have indicated this is useful for as a source of information on expected question flow and which questions are dependent.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -12,69 +12,69 @@ Rails.application.routes.draw do
 
     get "/session-expired", to: "session_expired#show"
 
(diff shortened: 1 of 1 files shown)


brucebolt (approved):
👍

PR #224, opened 2020-04-02
Set config.hosts for all environments

brucebolt:
If a user sets an X-Forwarded header in a request, this will in some cases override the Host header in a Rails app, resulting in URL redirection to a different website. If an attacker manages to poison a cache with such a header, users will be redirected to malicious sites.

In https://github.com/rails/rails/issues/29893, the resolution is to set config.hosts.  Therefore setting this for all environments in this app.

Trello card: https://trello.com/c/vdSUBzGw

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -4,6 +4,13 @@ Rails.application.configure do
   # Settings specified here will take precedence over those in
   # config/application.rb.
 
+  config.hosts = [
+    "coronavirus-vulnerable-people.service.gov.uk",
+    "d18j9d8kwes7fb.cloudfront.net",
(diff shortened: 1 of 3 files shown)


config/environments/production.rb
```diff
@@ -4,6 +4,12 @@
   # Settings specified here will take precedence over those in
   # config/application.rb.
 
+  config.hosts = [
+    "coronavirus-vulnerable-people.service.gov.uk",
+    "govuk-coronavirus-vulnerable-people-form-stg.cloudapps.digital",
+  ]
+
+
```
issyl0:
```suggestion
```

issyl0 (approved):
There's an extra blank line in the production file but it shouldn't block merging.

PR #227, opened 2020-04-03
Update gitignore

bilbof:
We can ignore these autogenerated assets.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,14 @@
 
 coverage
 tmp
+
+# ignore generated assets
+public/assets/component_guide
+public/assets/govuk_publishing_components
+public/assets/*.css
+public/assets/*.gz
+public/assets/*.js
+public/assets/*.json
+public/assets/*.png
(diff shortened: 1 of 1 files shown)


brucebolt (approved):

.gitignore
```diff
@@ -26,3 +26,14 @@
 
 coverage
 tmp
+
+# ignore generated assets
+public/assets/component_guide
+public/assets/govuk_publishing_components
+public/assets/*.css
```
alex-ju, line 33: It seems like we do have [one non-autogenerated asset in here](https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/blob/master/public/assets/error-pages.85f433bf6332a1245653b92b44d2430a55810be0.min.css) 😬

bilbof: 😬 I'll fix that.

alex-ju: That was quick. Thank you!

PR #229, opened 2020-04-03
Update NHS number page to match prototype

brucebolt:
Makes two changes that are present in the prototype but not the app:
- Removes the label for the input field
- Reduces width of the input field

Maintains the relationship of a label against the input field for accessibility reasons.

Before change:
Screenshot 2020-04-03 at 16 13 02

After change:
Screenshot 2020-04-03 at 16 12 35


alex-ju (approved):
👏

PR #230, opened 2020-04-03
Perform validations before setting session

brucebolt:
We are currently saving invalid responses to the session store before doing validations.  This means a user can submit an invalid response by browsing back to the "check your answers" page. 

To resolve this, we need to validate the response before updating the session store, but still replay the invalid response back to the user so they can correct it.  This changes the ordering of logic in the question controllers so we only update the session store values if validation has been successful.

Additionally, this PR:
- removes all use of session out of the views
- fixes two tests which were not correctly set up (but masked by the session hash being updated before validation took place)
- (as an unintentional side effect) resolves all string vs symbols issues for the stored session values, using symbols throughout now

Trello card: https://trello.com/c/PygFKz9a

### DIFF
--- a/app/controllers/coronavirus_form/live_in_england_controller.rb
+++ b/app/controllers/coronavirus_form/live_in_england_controller.rb
@@ -4,12 +4,13 @@ class CoronavirusForm::LiveInEnglandController < ApplicationController
   skip_before_action :check_first_question
 
   def submit
-    live_in_england = strip_tags(params[:live_in_england]).presence
-    session[:live_in_england] = live_in_england
+    @form_responses = {
+    live_in_england: strip_tags(params[:live_in_england]).presence,
+  }
 
     invalid_fields = validate_radio_field(
       controller_name,
-      radio: live_in_england,
+      radio: @form_responses[:live_in_england],
     )
 
     if invalid_fields.any?
@@ -19,11 +20,14 @@ class CoronavirusForm::LiveInEnglandController < ApplicationController
       respond_to do |format|
         format.html { render controller_path, status: :unprocessable_entity }
       end
-    elsif session[:live_in_england] == I18n.t("coronavirus_form.questions.live_in_england.options.option_no.label")
+    elsif @form_responses[:live_in_england] == I18n.t("coronavirus_form.questions.live_in_england.options.option_no.label")
+      session[:live_in_england] = @form_responses[:live_in_england]
       redirect_to not_eligible_england_url
     elsif session[:check_answers_seen]
+      session[:live_in_england] = @form_responses[:live_in_england]
       redirect_to check_your_answers_url
     else
+      session[:live_in_england] = @form_responses[:live_in_england]
       redirect_to nhs_letter_url
     end
   end
--- a/app/controllers/coronavirus_form/medical_conditions_controller.rb
+++ b/app/controllers/coronavirus_form/medical_conditions_controller.rb
@@ -2,12 +2,13 @@
 
 class CoronavirusForm::MedicalConditionsController < ApplicationController
   def submit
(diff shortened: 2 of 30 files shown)


bilbof (approved):
Excellent change 👏 🏅

app/controllers/coronavirus_form/live_in_england_controller.rb
```diff
@@ -4,12 +4,13 @@ class CoronavirusForm::LiveInEnglandController < ApplicationController
   skip_before_action :check_first_question
 
   def submit
-    live_in_england = strip_tags(params[:live_in_england]).presence
-    session[:live_in_england] = live_in_england
+    @form_responses = {
+    live_in_england: strip_tags(params[:live_in_england]).presence,
+  }
```
leenagupte, line 9: Nitpick: The indentation is off here.

app/controllers/coronavirus_form/live_in_england_controller.rb
```diff
@@ -19,11 +20,14 @@ def submit
...
       end
-    elsif session[:live_in_england] == I18n.t("coronavirus_form.questions.live_in_england.options.option_no.label")
+    elsif @form_responses[:live_in_england] == I18n.t("coronavirus_form.questions.live_in_england.options.option_no.label")
+      session[:live_in_england] = @form_responses[:live_in_england]
       redirect_to not_eligible_england_url
     elsif session[:check_answers_seen]
+      session[:live_in_england] = @form_responses[:live_in_england]
       redirect_to check_your_answers_url
     else
+      session[:live_in_england] = @form_responses[:live_in_england]
```
leenagupte, line 30: Seems odd to have this line repeated 3 times, but ok.

app/controllers/coronavirus_form/medical_conditions_controller.rb
```diff
@@ -2,12 +2,13 @@
 
 class CoronavirusForm::MedicalConditionsController < ApplicationController
   def submit
-    medical_conditions = strip_tags(params[:medical_conditions]).presence
-    session[:medical_conditions] = medical_conditions
+    @form_responses = {
+    medical_conditions: strip_tags(params[:medical_conditions]).presence,
+  }
```
leenagupte, line 7: Indentation is off here too. Looks like a copy and paste error.

app/helpers/field_validation_helper.rb
```diff
@@ -1,10 +1,10 @@
 # frozen_string_literal: true
 
 module FieldValidationHelper
-  def validate_mandatory_text_fields(mandatory_fields, page)
+  def validate_mandatory_text_fields(mandatory_fields, page, form_responses)
```
leenagupte, line 4: Where is this being used? I was expecting to see where it's called needing to be updated too. 🤔

leenagupte (commented):
Looks good! I love that we using all symbol keys now. Just a couple of nitpicks about the indentation.

app/controllers/coronavirus_form/live_in_england_controller.rb
brucebolt: Yeah, I struggled to find a way of only having this included once.  A separate `set_session_store` method seemed a bit of an overkill.

app/helpers/field_validation_helper.rb
brucebolt: Doesn't look like we are using it anywhere.  I'll remove it.

app/controllers/coronavirus_form/live_in_england_controller.rb
brucebolt: I'm surprised Rubocop didn't pick this up 😕.  I've raised https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/pull/234 to fix it.

PR #234, opened 2020-04-06
Fix issues from #230

brucebolt:
- Fixed indentation problem that rubocop did not pick up
- Removes unused `validate_mandatory_text_fields` method from field validation helper


leenagupte (approved):
👍

PR #237, opened 2020-04-08
Update Scotland link

StephenGill:
Update link for people in Scotland so it points to a page that's more specific.


issyl0 (approved):

issyl0 (approved):
Thanks, @StephenGill.

PR #239, opened 2020-04-09
Clarify what users should do after registering

StephenGill:
Updates text on the confirmation page to clarify what users should do after they have registered.


bilbof:
I fixed up the indentation and brought this patch over https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/pull/240

PR #242, opened 2020-04-09
Remove Postgres database service from PaaS manifest

issyl0:
- We use Dynamo in this app and we don't need to keep this Postgres db
  hanging around in Staging or Production.


richardTowers (approved):

issyl0:
I'll merge this later tonight as the commands needed to unbind the service and restart the app aren't zero-downtime, so I'll do this "out of hours".

PR #244, opened 2020-04-11
Use Github Actions for PR builds

huwd:
Trello: https://trello.com/c/r22AVt8W/207-switch-the-pr-builds-on-all-forms-to-use-github-actions-not-travis
Trello: https://trello.com/c/TNaiKhFV/228-make-travis-only-build-against-master

PR #250, opened 2020-04-13
ci/travis: Only run on master

issyl0:
- Now we've got GitHub Actions running for PR builds, we don't need to
  run Travis PR builds at the same time, so we can save a bit of the
  planet by only running the Travis checks on `master` (which we still
  need to do because the Concourse release process triggers on a
  successful Travis `master` build).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,6 @@
+branches:
+  only:
+  - master
 language: ruby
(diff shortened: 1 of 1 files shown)


issyl0:
Oh, wait, #244 isn't merged yet.

PR #251, opened 2020-04-14
Remove all whitespace from postcode

leenagupte:
Trello: https://trello.com/c/hUsoCjYe

## Motivation
Users sometimes enter their postcode with extra spaces, e.g "e1 8 qs"
which fails validation. This caused quite a few users to contact
support.

We need to update the validator so we accept valid postcodes that are
entered in the incorrect format, so users are not being rejected when
they enter a valid postcode.

### DIFF
--- a/app/controllers/coronavirus_form/support_address_controller.rb
+++ b/app/controllers/coronavirus_form/support_address_controller.rb
@@ -14,7 +14,7 @@ class CoronavirusForm::SupportAddressController < ApplicationController
         building_and_street_line_2: strip_tags(params[:building_and_street_line_2]&.strip).presence,
         town_city: strip_tags(params[:town_city]&.strip).presence,
         county: strip_tags(params[:county]&.strip).presence,
-        postcode: strip_tags(params[:postcode]&.strip).presence,
+        postcode: strip_tags(params[:postcode]&.gsub(/[[:space:]]+/, "")).presence,
       },
     }
 
--- a/spec/controllers/coronavirus_form/support_address_controller_spec.rb
+++ b/spec/controllers/coronavirus_form/support_address_controller_spec.rb
@@ -35,7 +35,7 @@ RSpec.describe CoronavirusForm::SupportAddressController, type: :controller do
         building_and_street_line_2: "1 Horse Guards Road",
         town_city: "London",
         county: "United Kingdom",
-        postcode: "SW1A 2HQ",
+        postcode: "SW1A2HQ",
       }
     end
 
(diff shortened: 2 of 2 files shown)


richardTowers:
Would it be overly paranoid to remove all non-alphanumeric characters, instead of just whitespace?

For example, on a mac `ALT+SPACE` will input a non-breaking-space, which doesn't count as whitespace in Ruby's regex thing, but would still break the validation.

brucebolt:
A concern I have with this approach is that we will replay the postcode on the "check your answers" page without a space (and also when editing a response).  Users may decide to then edit the response, only to find the space gets removed again (which could cause frustration if they end up in an editing loop).

Could we maybe use a more complex solution here: validate against a whitespace stripped version of the postcode, write this to the database but render the user entered value on the "check you answers" page and also when correcting response.  E.g.

```
    @form_responses = {
      support_address: {
...
        postcode_user: strip_tags(params[:postcode]&.strip).presence,
        postcode: strip_tags(params[:postcode]&.gsub(/\s+/, "")).presence,
      },
    }
```
In the contact details view we could use `postcode_user`, in check your answers use `postcode_user` but only write `postcode` to the database?

leenagupte:
> Would it be overly paranoid to remove all non-alphanumeric characters, instead of just whitespace?
> 
> For example, on a mac `ALT+SPACE` will input a non-breaking-space, which doesn't count as whitespace in Ruby's regex thing, but would still break the validation.

🤔 This might work: `.gsub(/[[:space:]]+/, "")`, "borrowed" from: https://markhneedham.com/blog/2013/02/23/ruby-stripping-out-a-non-breaking-space-character-nbsp/

```
[1] pry(main)> pc = "e1 8 qs"
=> "e1 8 qs"
[2] pry(main)> pc.gsub(/[[:space:]]+/, "")
=> "e18qs"
[3] pry(main)> pc = "e1 8 qs"
=> "e1 8 qs"
[4] pry(main)> pc.gsub(/[[:space:]]+/, "")
=> "e18qs"
[5] pry(main)> pc = "e1\u00A08\u00A0qs"
=> "e1 8 qs"
[6] pry(main)> pc.gsub(/[[:space:]]+/, "")
=> "e18qs"
```

richardTowers:
@tlwr pointed out that `str.gsub(/[[:space:]]/, '')` would get unicode spaces too ([example](https://rubular.com/r/yiWISuPB5HzhOv))

leenagupte:
> A concern I have with this approach is that we will replay the postcode on the "check your answers" page without a space (and also when editing a response). Users may decide to then edit the response, only to find the space gets removed again (which could cause frustration if they end up in an editing loop).
> 
> Could we maybe use a more complex solution here: validate against a whitespace stripped version of the postcode, write this to the database but render the user entered value on the "check you answers" page and also when correcting response. E.g.
> 
> ```
>     @form_responses = {
>       support_address: {
> ...
>         postcode_user: strip_tags(params[:postcode]&.strip).presence,
>         postcode: strip_tags(params[:postcode]&.gsub(/\s+/, "")).presence,
>       },
>     }
> ```
> 
> In the contact details view we could use `postcode_user`, in check your answers use `postcode_user` but only write `postcode` to the database?

Is it worth waiting to see if this is a problem first?

leenagupte:
@richardTowers @brucebolt I've updated the regex to use `[[:space:]]`

bilbof (approved):

brucebolt (approved):

PR #252, opened 2020-04-14
Add opt-in Google Analytics tracking

bilbof:
This adds Google Analytics to the form, to help us understand
how people are using the service to offer help, so the development
team can make improvements.

User tracking will only be enabled if the users opts in.

This uses the cookie banner component, and copies the approach
(cookie settings page) from GOV.UK for this service.

If a user opts in to usage tracking, we'll load Google Analytics
so we can see how users move through the form in aggregate, in
order to improve the application.

Every page will have a cookie banner until the user accepts or
rejects cookies.

## Screenshots

**Cookie page**

localhost_5000_cookies

**Banner**

Screenshot 2020-04-14 at 19 00 34

**Banner after accept**

Screenshot 2020-04-14 at 19 02 39

https://trello.com/c/ytRb9HIU/152

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -1,2 +1,9 @@
+//= require jquery
 // = require govuk/all.js
+//= require govuk_publishing_components/components/cookie-banner
+//= require govuk_publishing_components/lib/cookie-functions
+//= require analytics
+//= require cookies
 window.GOVUKFrontend.initAll()
+window.CookieSettings.start()
+window.GOVUK.analyticsInit()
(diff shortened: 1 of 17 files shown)


app/assets/javascripts/application.js
```diff
@@ -1,2 +1,9 @@
+//= require jquery
 // = require govuk/all.js
+//= require govuk_publishing_components/components/cookie-banner
+//= require govuk_publishing_components/lib/cookie-functions
+//= require analytics
+//= require cookies
+window.CookieSettings.start()
+window.GOVUK.analyticsInit()
```
alex-ju, line 9: Since these are not running asynchronously I would suggest initialising the UI components first and then the cookie settings and analytics.

alex-ju (approved):
Looks good. I left a comment regarding the initialisation order, otherwise it's good to go.

PR #255, opened 2020-04-16
ci: Optimise Actions runs by caching `bundle install`

issyl0:
The "build app" portion of our builds take four minutes every time.
This is a lot slower than Travis. So we can cache bundle install and
see if that improves things.


huwd (approved):

bilbof (approved):
💥 🚀

PR #257, opened 2020-04-16
Rename db in development and test

leenagupte:
The vulnerable people, business volunteer and find support forms all use the same
database names in development and testing. This is fine in production where they're
all hosted on different RDS instances, but can produce some interesting results locally.

Need to run `bundle exec rails dynamoid:create_tables` to setup the db again.


leenagupte:
Shouldn't have been closed.

PR #266, opened 2020-04-17
Authenticate before before_actions

bilbof:
This fixes an issue where we would attempt to redirect a user
prior to authenticating them. This had the effect of making
them enter their basic auth credentials twice.

This broke the smoke tests when testing redirects to the first
question were answered because the before_action that redirected
a user to the first page when they hadn't answered it came before
we had authenticated their initial request.

Co-Authored-By: Graham Lewis


gclssvglx (approved):
LGTM 👍

PR #268, opened 2020-04-17
Remove Travis 

Rosa-Fox:
[Trello](https://trello.com/c/t6jxzyCY/229-dockerize-app-and-run-tests-within-concourse)

- Removed Travis
- Added `run-quality-checks` into the pipeline
- Simplified some of the naming within the pipeline
- Added Dynamo DB as a Docker dependency for tests

Co-authored-by: Huw Diprose <huw.diprose@digital.cabinet-office.gov.uk>

vulnerable-people-pipeline

### DIFF
--- a/concourse/Dockerfile
+++ b/concourse/Dockerfile
@@ -13,7 +13,6 @@ RUN apt-get update --fix-missing && apt-get -y upgrade \
     && apt-get install -y --no-install-recommends \
       google-chrome-unstable \
       default-jre \
-      --no-install-recommends \
     && rm -rf /var/lib/apt/lists/* \
     && rm -rf /src/*.deb
 
--- a/concourse/Dockerfile
+++ b/concourse/Dockerfile
@@ -16,12 +16,12 @@ RUN apt-get update --fix-missing && apt-get -y upgrade \
 
 ENV CHROME_NO_SANDBOX=true
 
-COPY Gemfile* .ruby-version /home/chrome/src/
+COPY Gemfile* .ruby-version /application/
 
-WORKDIR /home/chrome/src
+WORKDIR /application/
 
 RUN bundle install
 
-COPY . /home/chrome/src/
+COPY . /application/
 
 ENTRYPOINT ["/bin/bash"]
--- a/concourse/Dockerfile
+++ b/concourse/Dockerfile
@@ -10,9 +10,13 @@ RUN apt-get update --fix-missing && apt-get -y upgrade \
     && wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add - \
     && sh -c 'echo "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google.list' \
     && apt-get update \
-    && apt-get install -y google-chrome-unstable --no-install-recommends \
+    && apt-get install -y --no-install-recommends \
+      google-chrome-unstable \
+      default-jre \
+      --no-install-recommends \
     && rm -rf /var/lib/apt/lists/* \
     && rm -rf /src/*.deb
+
 
 ENV CHROME_NO_SANDBOX=true
 
--- a/concourse/pipeline.yml
+++ b/concourse/pipeline.yml
@@ -5,81 +5,65 @@ resource_types:
     source:
       repository: nulldriver/cf-cli-resource
 
-  - name: travis
-    type: docker-image
-    source:
-      repository: rbakergds/travis-resource
-      tag: latest
-
 resources:
-  - name: govuk-coronavirus-vulnerable-people-form
+  - name: git-master
     type: git
     icon: github-circle
     source:
       uri: https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form
       branch: master
 
-  - name: govuk-coronavirus-vulnerable-people-form-feature-tests
+  - name: git-master-gems-dockerfile
     type: git
     icon: github-circle
     source:
       uri: https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form
       branch: master
       paths:
+        - concourse/Dockerfile
         - Gemfile*
-        - .ruby-version
-        - spec/
 
   - name: every-day
     type: time
     source:
       interval: 24h
 
-  - name: govuk-coronavirus-vulnerable-people-form-travis-build
-    type: travis
-    icon: sync
-    source:
-      repository: alphagov/govuk-coronavirus-vulnerable-people-form
-      travis-token: ((travis-api-token))
-      branch: master
-      pro: true
-
-  - name: feature-tests-image
+  - name: tests-image
     type: docker-image
     icon: docker
     source:
       repository: ((readonly_private_ecr_repo_url))
-      tag: govuk-coronavirus-vulnerable-people-form-feature-tests
+      tag: govuk-coronavirus-vulnerable-people-tests-image
 
 jobs:
   - name: update-pipeline
     plan:
-      - get: govuk-coronavirus-vulnerable-people-form
+      - get: git-master
         trigger: true
       - set_pipeline: govuk-corona-vulnerable-people-form
-        file: govuk-coronavirus-vulnerable-people-form/concourse/pipeline.yml
+        file: git-master/concourse/pipeline.yml
 
   - name: build-feature-tests-image
     serial: true
     plan:
       - get: every-day
         trigger: true
-      - get: govuk-coronavirus-vulnerable-people-form-feature-tests
+      - get: git-master-gems-dockerfile
         trigger: true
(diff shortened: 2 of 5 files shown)


concourse/Dockerfile
```diff
@@ -10,18 +10,22 @@ RUN apt-get update --fix-missing && apt-get -y upgrade \
     && wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add - \
     && sh -c 'echo "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google.list' \
     && apt-get update \
-    && apt-get install -y google-chrome-unstable --no-install-recommends \
+    && apt-get install -y --no-install-recommends \
+      google-chrome-unstable \
+      default-jre \
```
huwd, line 15: Never sure what the best version of java to install is, this seems to be the most lightweight one that gets us a `java` command we can use.

Shout if there's a better one?

concourse/Dockerfile
```diff
@@ -10,18 +10,22 @@ RUN apt-get update --fix-missing && apt-get -y upgrade \
...
+      --no-install-recommends \
```
huwd:
```suggestion
```
oops, we missed that, we can remove this line, it's now on Line 13

concourse/pipeline.yml
```diff
@@ -5,117 +5,133 @@ resource_types:
...
+          platform: linux
+          run:
+            dir: git-master
+            path: bash
+            args:
+              - -c
+              - |
+                set -eu
+                wget https://s3.eu-central-1.amazonaws.com/dynamodb-local-frankfurt/dynamodb_local_latest.tar.gz
+                sha256sum -c <<< "d77f5371c21868dd489b53782619b7f2e418d4b1068947a28d80082052ca1d64 dynamodb_local_latest.tar.gz"
```
huwd, line 81: So... this was my stupid idea. This is extremely fragile and will break the moment amazon update the "latest" dynamo version.

I really don't like the idea of downloading it without doing a checksum check though...

they do publish the checksums [alongside here ](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DynamoDBLocal.DownloadingAndRunning.html), I just can't for the life of me work out how to do an automated comparison that doesn't seem like terrible command-line-fu.

Any cli blackbelts around to help?

huwd: Alternatively we stick this in some docs and it's a feature whenever it breaks because a Dev can check for anything naferious?

huwd: Or we tell it to download something other than latest, like a specific version, which we then stick to until we need to upgrade? Would be a good idea to match the AWS version?

concourse/pipeline.yml
```diff
@@ -5,117 +5,133 @@ resource_types:
...
-          CF_STARTUP_TIMEOUT: 5 # minutes
-          REQUIRE_BASIC_AUTH: "true"
-          BASIC_AUTH_PASSWORD: ((basic-auth-password))
-          AWS_ACCESS_KEY_ID: ((aws-access-key-id))
-          AWS_SECRET_ACCESS_KEY: ((aws-secret-access-key))
-          SECRET_KEY_BASE: ((secret-key-base-staging))
-          AWS_DYNAMODB_SUBMISSIONS_TABLE_NAME: coronavirus-vulnerable-people-staging
-          AWS_ASSETS_BUCKET_NAME: govuk-coronavirus-vulnerable-people-assets-staging
-          HOSTNAME: govuk-coronavirus-vulnerable-people-form-stg
+          file: git-master/concourse/tasks/deploy-to-govuk-paas.yml
```
huwd: Ah i think we borked the indentation here @Rosa-Fox we can tweak this on monday?

concourse/pipeline.yml
issyl0: Yeah, I'd much prefer your third suggestion.

Other places in `org:alphagov` (GitHub search) that check the sha256sum of a file do it the way you've done it here.

concourse/pipeline.yml
```diff
@@ -5,117 +5,133 @@ resource_types:
...
+            INSTANCES: 5
+            CF_STARTUP_TIMEOUT: 5 # minutes
+            REQUIRE_BASIC_AUTH: "true"
+            BASIC_AUTH_PASSWORD: ((basic-auth-password))
+            AWS_ACCESS_KEY_ID: ((aws-access-key-id))
+            AWS_SECRET_ACCESS_KEY: ((aws-secret-access-key))
+            SECRET_KEY_BASE: ((secret-key-base-staging))
+            AWS_DYNAMODB_SUBMISSIONS_TABLE_NAME: coronavirus-vulnerable-people-staging
+            AWS_ASSETS_BUCKET_NAME: govuk-coronavirus-vulnerable-people-assets-staging
+            HOSTNAME: govuk-coronavirus-vulnerable-people-form-stg
```
issyl0:
```suggestion
        file: git-master/concourse/tasks/deploy-to-govuk-paas.yml
        params:
          CF_SPACE: staging
          INSTANCES: 5
          CF_STARTUP_TIMEOUT: 5 # minutes
          REQUIRE_BASIC_AUTH: "true"
          BASIC_AUTH_PASSWORD: ((basic-auth-password))
          AWS_ACCESS_KEY_ID: ((aws-access-key-id))
          AWS_SECRET_ACCESS_KEY: ((aws-secret-access-key))
          SECRET_KEY_BASE: ((secret-key-base-staging))
          AWS_DYNAMODB_SUBMISSIONS_TABLE_NAME: coronavirus-vulnerable-people-staging
          AWS_ASSETS_BUCKET_NAME: govuk-coronavirus-vulnerable-people-assets-staging
          HOSTNAME: govuk-coronavirus-vulnerable-people-form-stg
```

concourse/pipeline.yml
issyl0: Fixed it for you Huw!

concourse/pipeline.yml
```diff
@@ -5,117 +5,133 @@ resource_types:
...
   - name: deploy-to-staging
     serial: true
     plan:
-      - get: govuk-coronavirus-vulnerable-people-form-travis-build
+      - get: git-master
         trigger: true
-      - get: govuk-coronavirus-vulnerable-people-form
+        passed: [run-quality-checks]
       - task: deploy-to-paas
         config:
```
huwd, line 95: We indented this properly, but then the concourse validator started complaining file: was invalid.

We've got this on a few other pipelines. I'm not sure what the syntax or indentation should be. seems really wrong to have a `config:` on its own line and everything else at the same level too.

Might need to also check this on:
- [find support](https://github.com/alphagov/govuk-coronavirus-find-support/blob/master/concourse/pipeline.yml#L93-L105)
- [biz form](https://github.com/alphagov/govuk-coronavirus-business-volunteer-form/blob/master/concourse/pipeline.yml#L75-L96)

comparing them i think it's a redundant key that's just rendering to `nil`. It was only there because we had some copy paste stuff from some verify concourse pipelines. I think it can be removed... and should be elsewhere too!

concourse/pipeline.yml
huwd: that's a relief, then we weren't totally insane going down this route.
Spent ages reading "how-to" guides that were all like... check it manually...

huwd: looks like they have all the versions: https://s3.eu-central-1.amazonaws.com/dynamodb-local-frankfurt/

concourse/pipeline.yml
issyl0: What's "the Concourse validator"?

concourse/Dockerfile
alex-ju: `default-jre` is perfect, it will give you the latest stable `openjdk-#-jre` depending on distribution version

huwd: Thanks alex!

concourse/pipeline.yml
huwd: When you run `set-pipeline` in fly it takes in the provided YAML and runs some validation checks. I'm pretty sure it's the same thing as if you run `validate-pipeline` [see here](https://concourse-ci.org/setting-pipelines.html#fly-validate-pipeline). It'll tell you if you have any obvious syntax errors in your `pipeline.yml`.

I've also had an odd case where it made it past the validatior, but then i guess something in concourse compiles it into JSON. Got a really unhelpful "JSON is invalid" kind-of error without a line number. Turned out I'd indented something wrong in the yml. So i guess that's another kind of validation...

issyl0: Ohhh, right, yes!

bilbof (approved):
LGTM

huwd:
We could merge this now, or sort the checksum thing first.
I'm currently just trying to get into AWS and see if I can find what version we're using

PR #271, opened 2020-04-20
Use I18n lookup rather than hard-coded strings in tests

leenagupte:
Trello: https://trello.com/c/aZdkhCC0

This should make it easier for content designers and others to change content without worrying about breaking tests.


issyl0 (approved):
😍

PR #273, opened 2020-04-21
Reject date of birth older than 150 years

brucebolt:
Shows an error message if a user enters a date of birth that is more than 150 years into the past.  This will reduce validation errors when the form responses are triaged.

Screenshot 2020-04-21 at 13 47 17

Trello card: https://trello.com/c/4uOY1kyH

### DIFF
--- a/app/helpers/field_validation_helper.rb
+++ b/app/helpers/field_validation_helper.rb
@@ -76,6 +76,10 @@ module FieldValidationHelper
     unless(invalid_fields != [] || DateTime.new(year.to_i, month.to_i, day.to_i) < Time.zone.now)
       invalid_fields << { field: "#{field}-day", text: t("coronavirus_form.errors.date_order") }
     end
+    # Check for date being more than 150 years ago
+    unless(invalid_fields != [] || DateTime.new(year.to_i, month.to_i, day.to_i) >= 150.years.ago)
+      invalid_fields << { field: "#{field}-year", text: t("coronavirus_form.errors.invalid_date") }
+    end
     invalid_fields
   end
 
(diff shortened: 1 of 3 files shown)


app/helpers/field_validation_helper.rb
```diff
@@ -76,6 +76,10 @@ def validate_date_of_birth(year, month, day, field)
     unless(invalid_fields != [] || DateTime.new(year.to_i, month.to_i, day.to_i) < Time.zone.now)
       invalid_fields << { field: "#{field}-day", text: t("coronavirus_form.errors.date_order") }
     end
+    # Check for date being more than 150 years ago
+    unless(invalid_fields != [] || DateTime.new(year.to_i, month.to_i, day.to_i) >= Time.zone.now - 150.years)
```
leenagupte: Why the check for invalid_fields not being blank? I realise that it's the same as the other checks, but I'm curious.

JonathanHallam (approved):
Looks good to me :)

app/helpers/field_validation_helper.rb
brucebolt: You could end up with a list of validation errors with (potentially) all the same error message if every validation were to run

leenagupte: Also, `Time.zone.now - 150.years` could be written as `150.years.ago`

brucebolt: Thanks, have updated to use `150.years.ago`

leenagupte: Perhaps rather than adding `invalid_fields != []` to the beginning of each check, they can all be wrapped in:
```
if invalid_fields.blank?
   date_off_birth = DateTime.new(year.to_i, month.to_i, day.to_i)
   
  if date_off_birth in the future
  if date_off_birth >= 150.years.ago
end
```
etc

bilbof (approved):
Nice 👍

app/helpers/field_validation_helper.rb
brucebolt: Perhaps we should consider this in another refactor.  I'm sure there was a design reason why we're showing a single error message, rather than doing all the validations at once.

PR #277, opened 2020-04-22
Improve email address validation

issyl0:
- We were previously only validating that the email address contains the
  `@` character. That's not enough, as people can enter invalid email
  addresses, like ones with commas in, if they mistype `.com` or
  somesuch.
- Our downstream data consumers were finding this out the hard way when
  they tried to use the email addresses to contact people.
- This uses the Python implementation at https://emailregex.com/ to
  validate the email address doesn't contain any invalid characters.

https://trello.com/c/hGSi37CH/315-email-address-field-accepts-commas

### DIFF
--- a/app/helpers/field_validation_helper.rb
+++ b/app/helpers/field_validation_helper.rb
@@ -84,7 +84,7 @@ module FieldValidationHelper
   end
 
   def validate_email_address(field, email_address)
-    if email_address =~ /@/
+    if email_address =~ /(^[a-zA-Z0-9_.+\-]+@[a-zA-Z0-9\-]+\.[a-zA-Z0-9\-.]+$)/
       []
     else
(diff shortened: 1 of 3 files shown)


bilbof (changes requested):
If possible could we have some more tests of this helper, with various email combinations? Just want to make sure we have various examples of invalid/valid emails in the test cases.

bilbof (approved):
🚀 awesome, thank you so much

issyl0:
Thanks for the review, Bill. We can't merge this until we've heard back from DWP.

leenagupte (dismissed):
Needs stakeholder approval.

huwd:
@issyl0 [Steve has confirmed ](https://trello.com/c/hGSi37CH/315-email-address-field-accepts-commas)this can be merged without waiting for a reply

So looks like we can resolve that conflict and :rocket:

PR #286, opened 2020-04-24
Add link to session expired page

JonathanHallam:
This warns the user that they might be seeing that page because they don't have cookies enabled. It also provides a link to remedy this.


alex-ju (approved):
I suspect `form_with_js.feature` is part of another change and not needed here.

JonathanHallam:
@alex-ju yup, sorry, no idea how I missed that!

JonathanHallam:
@alex-ju fixed!

PR #288, opened 2020-04-24
WIP: allow user to be referred from another application

brucebolt:
This allows them to skip questions they would have already answered in
the find support (triage) tool

What
----

Describe what you have changed and why.

How to review
-------------

Describe the steps required to review and test the changes.

Links
-----

Add links to any relevant [Trello cards](https://trello.com/b/9Pvq5z2n/govuk-coronavirus-services-team-doing-board),
Zendesk tickets or Sentry issues.

### DIFF
--- a/app/controllers/coronavirus_form/refer_from_other_app_controller.rb
+++ b/app/controllers/coronavirus_form/refer_from_other_app_controller.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+class CoronavirusForm::ReferFromOtherAppController < ApplicationController
+  def from_find_support_app
+    ## Sets the session variables that the user would have already answered in Support app
+    session[:live_in_england] = I18n.t("coronavirus_form.questions.live_in_england.options.option_yes.label")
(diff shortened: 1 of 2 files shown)


app/controllers/coronavirus_form/refer_from_other_app_controller.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class CoronavirusForm::ReferFromOtherAppController < ApplicationController
+  def from_find_support_app
```
bilbof, line 4: This is a good solution 👍

If one app changes we might need to update the other. If we see this happening, we could rename the methods / controller to be more generic, eg

```
class CoronavirusForm::SkipQuestionsController < ApplicationController
  def skip_questions
```

`skip_questions` would then populate the session given some params and redirect to the correct URL. I don't think that's necessary to do now though.

bilbof:
We can close this since the spike was completed successfully, and we no longer need to implement this feature.

PR #291, opened 2020-04-27
Always store date of birth in the same format

brucebolt:
What
----

We have found some users enter date of births in differing formats, which can cause confusion, or errors, in the downstream processing.

Therefore ensuring that the year is always stored with 4 digits (e.g. to avoid confusion whether 01 is 1901 or 2001), the month and day will always be two digits.

How to review
-------------

App link: https://coronavirus-improve-dob-fatgec.herokuapp.com/live-in-england

- Review code.
- Run app locally, or in Heroku review.  Navigate to `/date-of-birth`.  Try various dates, but specifically those with a single digit for the day or month.  Then go to `/check-your-answers`.  You should see the dates padded with a leading zero, which will be the format we save the data.

Links
-----

Trello card: https://trello.com/c/S5XTLtzH


bilbof (approved):
🎉

PR #293, opened 2020-04-28
Backend integration with postcode API

gclssvglx:
```
GIVEN I have an address
THEN we will need to associate it with a UPRN
```

Local authorities will need to match these against their own databases and they need UPRNs for this - we can't infer these from postcodes alone as UPRNs are residence specific.

We need a way for a specific UPRN to be selected from the list of available addresses that are associated with a postcode.

This change adds a backend API call to the Ordnance Survey Places API which returns a list of UPRNs and specific address for a given postcode.

HTTP tests are done either against a [VCR](https://github.com/vcr/vcr) cassette or stubbed using webmock.  Neither of these require 'live' tests therefore.

As a consequence of adding this API an API key is required for production use.  An API key can be obtained by first [registering here](https://developer.ordnancesurvey.co.uk/) then generate an API key for the [Places API](https://developer.ordnancesurvey.co.uk/os-places-api) in your profile.  An API key is also required if you need to change or regenerate the existing VCR cassettes.  To do this, ensure you have a valid API set in your local environment called `ORDNANCE_SURVEY_PLACES_API_KEY` then simply delete the `address` directory under `ROOT/spec/vcr` and re-run the tests as normal.

Trello: https://trello.com/c/V8pQIb8z/333-backend-integration-with-postcode-api


app/helpers/address_helper.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module AddressHelper
+  def postcode_lookup(postcode)
+    postcode_url = "#{ENV['API_URL']}?postcode=#{postcode}&key=#{ENV['API_KEY']}"
```
brucebolt: One thing we discovered was that the default response excludes flats that do not have their own street address.  They can be included by adding `dataset=LPI` to the query string.

app/helpers/address_helper.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    response = Typhoeus.get(postcode_url)
+
+    results = []
+    if response.code == 200
+      addresses = JSON.parse(response.body)["results"]
+      addresses.each do |address|
+        results << { address.dig("DPA", "UPRN") => address.dig("DPA", "ADDRESS") }
```
brucebolt: If you make my proposed change to add `dataset=LPI` to the query string, then `DPA` here will need changing to `LPI`.

app/helpers/address_helper.rb
brucebolt: Would it be possible to make the environment variable names a little more descriptive, so we know what they are for when looking in the PaaS console?  E.g. `OS_PLACES_API_URL` and `OS_PLACES_API_KEY`.

.gitignore
```diff
@@ -37,3 +37,6 @@ public/assets/*.json
 
 # don't ignore these css
 !public/assets/error-pages.85f433bf6332a1245653b92b44d2430a55810be0.min.css
+
+# Dotenv
+.env
```
brucebolt: I don't think we'll need to use a `.env` file, or use the `dotenv-rails` gem.  GOV.UK PaaS lets us define environment variables (and therefore secrets) for each application: https://docs.cloud.service.gov.uk/deploying_apps.html#environment-variables.  We should be able to get a key for staging and another for production, then add them using the CloudFoundry CLI.

config/locales/en.yml
```diff
@@ -188,6 +188,12 @@ en:
 
       <h2 class="govuk-heading-m">If you didn’t expect to see this page</h2>
       <p class="govuk-body">You may be seeing this page because you’ve turned off cookies in your browser. You’ll need to <a class="govuk-link" href="https://ico.org.uk/your-data-matters/online/cookies/" rel="external" target="_blank">turn cookies on</a> before you can use this service.</p>
+  address_api:
```
brucebolt: Given these won't be displayed to the user at any point, do we need to store these as an internationalisation?  The English text of a status code can be obtained using (e.g.) `Rack::Utils::HTTP_STATUS_CODES[405]`.

app/helpers/address_helper.rb
```diff
@@ -0,0 +1,21 @@
...
+      end
+    else
+      results << { response.code => I18n.t("address_api.error_#{response.code}") }
+    end
+
+    { "code" => response.code, "results" => results }.to_json
```
brucebolt: I think returning a Rails hash here would be acceptable, since the helper will only be used within this application, so we'd always have to parse the JSON in every use.

spec/helpers/address_helper_spec.rb
```diff
@@ -0,0 +1,119 @@
...
+          response = JSON.parse(helper.postcode_lookup("SW1P2LQ"))
+
+          expect(response["code"]).to be(200)
+          expect(response["results"].count).to be(14)
+        end
+      end
+
+      it "returns a single UPRN for an address with it's own postcode" do
+        VCR.use_cassette "address/200_single_response" do
+          response = JSON.parse(helper.postcode_lookup("SW1A2AA"))
```
brucebolt: If we return a hash in https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/pull/293/files#diff-d0becce4dbc49e088754c6a09e42ca4eR19 rather than a JSON string, then we won't need to parse the JSON on each request.

brucebolt (commented):
This is looking good, great work implementing this.

I haven't had the time to do a thorough review, but I've left a few comments after a scan through your code.  The main thing is a very minor fix but resolves a significant issue we experienced during the spike where some flats (those without their own street number, but only some cases, e.g. houses converted to flats) were not being included in the default response.

I don't think we use VCR for stubbing requests elsewhere in GOV.UK, but it is definitely used elsewhere in GDS.  Looks like an interesting way of doing this.  Maybe we should have a short section of dev guidance on how to generate/update these stubs?

app/helpers/address_helper.rb
gclssvglx: Thanks @brucebolt - I've updated this to `ORDNANCE_SURVEY_PLACES_API_KEY`

config/locales/en.yml
gclssvglx: Thanks @brucebolt - at the moment, my working assumption is that we might want to display suitable/understandable (content provided) error messages to the user.  If that turns out an incorrect assumption then it makes sense to either use these strings or those provied by the OS Places API.  I would add that OS have changed the meaning of a few of the normal HTTP error codes slightly - so using theirs might be more consistent.  So, for now - these are simply stub text, pending the show/no-show UI decision.

app/helpers/address_helper.rb
gclssvglx: Thanks @brucebolt - I've update as per your suggestion.

app/helpers/address_helper.rb
gclssvglx: Thanks @brucebolt - I've update as per your suggestion.

.gitignore
gclssvglx: Thanks @brucebolt - Yes, quite right - `dotenv` isn't required, so has been removed.

app/helpers/address_helper.rb
gclssvglx: Thanks @brucebolt - I had initially thought that I'd be implementing all the way onto the UI (now a separate card) - hence the JSON response.  But, absolutely right that is not needed as part of the this card.  Changed as per your suggestion.

spec/helpers/address_helper_spec.rb
gclssvglx: Thanks @brucebolt - updated as per your suggestion.

gclssvglx:
@brucebolt Thanks very much for your review and comments 🥇.  I've implemented all but one - the I18N comment (see response above).  I've also updated a few things - naming mostly and added a VCR test for 500 (which should be a 405!), now I know how to generate one.  All the best 👍

huwd:
Interesting CI Tests are getting  `VCR::Errors::UnhandledHTTPRequestError:`

Gemfile
```diff
@@ -12,21 +12,26 @@ gem "dynamoid"
 gem "fog-aws"
 gem "govuk_app_config", "~> 2.1.1"
 gem "govuk_publishing_components", "~> 21.42.0"
+gem "json"
```
bilbof: You could remove this line I think, since `json` is a [default Ruby gem](https://stdgems.org/json/).

Gemfile
```diff
@@ -12,21 +12,26 @@ gem "dynamoid"
...
 gem "sentry-raven", "~> 3.0"
 gem "telephone_number", "~> 1.4"
 gem "timecop"
+gem "typhoeus"
 gem "uglifier", "~> 4.2"
 
 group :development do
   gem "listen", "~> 3"
+  gem "vcr"
+  gem "webmock"
```
bilbof: Should these be moved into the `:test` group?

app/helpers/address_helper.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    results = []
+    if response.code == 200
+      addresses = JSON.parse(response.body)["results"]
+      addresses.each do |address|
+        results << { address.dig("LPI", "UPRN") => address.dig("LPI", "ADDRESS") }
+      end
+    else
+      results << { response.code => I18n.t("address_api.error_#{response.code}") }
+    end
```
bilbof: Currently this method returns `[{ 500 => "Error 500" }]` or `[{ "uprn_1" => "10 Downing St..." }, ...]`. When we use this, we must check if any of the keys are non-200 status codes, which seems like unnecessary indirection.

I wouldn't be afraid to create new classes if it helps here. You might find this interface easier to express if you used a class instead of a module, since it could have state, and you could write methods that use that state e.g. `address_response.ok?` and `address_response.addresses`.

For example

```ruby
class AddressLookup
    def candidate_addresses
        @candidate_addresses ||= begin
        postcode_url = "#{API_URL}&postcode=#{postcode}&key=#{Rails.application.secrets.ordinance_survey_api_key}"
        response = Typhoeus.get(postcode_url)
        raise AddressLookupError unless response.status == 200
        present_addresses response
        end
    end

private

    API_URL = "".freeze

    def present_addresses(response)
        JSON.parse(response.body).fetch("results", []).map { ... }
    end
end
```

Gemfile
gclssvglx: Hi @bilbof - just adding a bit of context on this.  You're absolutely, right `json` is a default gem and shouldn't need to be added here.  However, without this in the Gemfile - I found I was getting some odd warning messages when running the tests.  These go away with it's inclusion :-/

config/locales/en.yml
```diff
@@ -188,6 +188,12 @@ en:
...
+    error_400: "Bad Request"
+    error_401: "Unauthorized"
+    error_404: "Not Found"
+    error_405: "Method Not Allowed"
+    error_500: "Internal Server Error"
```
bilbof: I don't expect the class that uses the address lookup class to use this text, so I'm not sure this is necessary in its current form.

From the perspective of someone consuming this class, we'd want the address lookup class to raise an error, or have a method like `.succeeded?` or something like that, rather than returning text.

spec/helpers/address_helper_spec.rb
```diff
@@ -0,0 +1,121 @@
+require "spec_helper"
+
+RSpec.describe AddressHelper, type: :helper do
+  let(:api_url) { AddressHelper.api_url }
+
+  context "allow net connections" do
+    before do
+      WebMock.allow_net_connect!
```
bilbof: We shouldn't need to allow real connections in test. Is it possible to remove this line, now that we've obtained the expected responses we can test against?

bilbof: I also wonder whether VCR might be overkill for this situation. Since you've installed webmock, one could do..

```ruby
stub_request(:get, "<postcode url>").
  with(body: successful_address_lookup_fixture, status: 200)
```

bilbof (changes requested):
I added a few comments inline.

I think it would also be helpful to introduce a retry mechanism into this class. I know in HTTParty one can rescue [a gem error](https://github.com/jnunemaker/httparty/blob/master/lib/httparty/exceptions.rb#L3) and retry using that, I wonder if Typhoeus has something similar?

Gemfile
bilbof: > These go away with it's inclusion

You might need to `require 'json'` in the files where you use `json`, though I'm not sure what warnings you were seeing.

spec/helpers/address_helper_spec.rb
leenagupte: I understand why Graham chose to use VCR. I believe that the first time we run the tests we're doing an actual call to the endpoint, hence allowing net connections, and are recording the response, so that should mean that the fixture files are accurate, and that we're not guessing what the response looks like.

pixeltrix: You should be able to hook directly into WebMock with VCR like this in spec/support/vcr_config.rb:

``` ruby
VCR.configure do |config|
   config.cassette_library_dir = "spec/vcr"
   config.hook_into :webmock
end
```

That will handle the allow/disable net connect requirements [automatically][1].

[1]: https://github.com/vcr/vcr/blob/master/lib/vcr/library_hooks/webmock.rb

pixeltrix:
Was there a reason to use Typhoeus instead of one of the other two HTTP libraries already in the Gemfile (excon and faraday)? Seems like it's more suitable for speeding up multiple concurrent requests which isn't the use case here as far as I can see here plus there's the added dependency on libcurl.

On Petitions we call the Parliament Data API a lot for postcode to constituency lookup and we found that maintaining a persistent HTTP connection to the API with a cached per-thread client offered us the best performance as can be seen in alphagov/e-petitions#380. Admittedly I didn't test Typhoeus at the time but most of the performance gain comes from not having to negotiate a new HTTP connection on every request and so it seems like the same logic would apply here.

It may be the case that you don't need absolute performance in which case I'd still recommend just using Faraday with the standard net/http adapter which is already included in Gemfile.lock by the sentry-raven gem - it has similar simple request methods like `Faraday.get` and can be configured to raise an exception on a HTTP error which makes handling them with rescue clauses much cleaner.

config/locales/en.yml
leenagupte: They were only being used in the tests so they've been removed. I expect that when the consumer of AddressLookup is implemented, a decision will be made about whether these entries are needed.

bilbof (approved):
Looks good :shipit: 🎉

leenagupte:
@bilbof, @brucebolt, Thanks for your reviews. I've made some changes since then. Most notably, I've followed @pixeltrix's suggestion, and am using Faraday rather than Typhoeus as the HTTP client.

Also I've tidied up the commit history, and made a stab at adding some documentation (this will need to be updated by people who know more). 

Could you take another look?

cc @gclssvglx

bilbof (approved):
Awesome! And thanks for adding docs, those look really helpful.

pixeltrix:
@leenagupte I suggestion I'd add is to make the dependency on Faraday explicit in the Gemfile - don't want someone removing Sentry and then it disappears from the Gemfile.lock.

leenagupte:
> @leenagupte I suggestion I'd add is to make the dependency on Faraday explicit in the Gemfile - don't want someone removing Sentry and then it disappears from the Gemfile.lock.

Thanks, I've added a PR to do this here: https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/pull/304

PR #298, opened 2020-04-30
Update privacy notice

StephenGill:
Update privacy notice as requested by privacy team


issyl0:
I labelled this PR manually as the label action doesn't work on forked PRs, because it needs permissions to write to the repo, which forks aren't given.

issyl0 (approved):

PR #304, opened 2020-05-04
Make the dependency on Faraday explicit

leenagupte:
Follows on from: https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/pull/293

Faraday was added as dependency of "sentry-raven", but PR #293 we started to use it directly to query the Ordnance Survey API.

This makes sure that we don't lose Faraday if sentry is ever removed.


Gemfile
```diff
@@ -9,6 +9,7 @@ gem "rails", "~> 6.0.2"
 gem "asset_sync"
 gem "bootsnap", "~> 1"
 gem "dynamoid"
+gem "faraday"
```
bilbof: I guess we should add a version for this, I'm not sure why the asset-sync / dynamoid ones don't have versions set.

leenagupte: Updated and fixed up.

bilbof (approved):

PR #308, opened 2020-05-05
Send confirmation SMS on registration

brucebolt:
What
----

Sends a confirmation text message to registrants who have entered a UK mobile telephone number into the "phone number for text messages" field.

Content is currently awaiting departmental sign-off, so this PR cannot yet be merged.

Example of SMS:
kAJUmrd8

How to review
-------------

- Review code changes.
- Set up and run application locally using a Notify API token.  Ensure text message is sent through Notify interface.

Links
-----

Trello card: https://trello.com/c/8L0b1iIg

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ You'll need to pass a GOV.UK Notify API key as an environment variable
 config.action_mailer.delivery_method = :notify
 ```
 
-You'll also need to set a `GOVUK_NOTIFY_TEMPLATE_ID`, which might involve
+You'll also need to set a `GOVUK_NOTIFY_EMAIL_TEMPLATE_ID`, which might involve
 creating a template in Notify if your Notify service doesn't have one.
 
 The template should have a Message of `((body))` only.
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ You'll need to pass a GOV.UK Notify API key as an environment variable
 `NOTIFY_API_KEY`, and change the delivery method in [development.rb][]:
 
 ```ruby
-config.action_mailer.delivery_method = :notify
+config.action_mailer.delivery_method = :notify_email
 ```
 
 You'll also need to set a `GOVUK_NOTIFY_EMAIL_TEMPLATE_ID`, which might involve
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ to process the email queue.
 Sidekiq will start automatically when you run `foreman start`, but you can
 also run it alone with `bundle exec sidekiq`.
 
-#### Sending emails locally
+#### Sending emails or SMSs locally
 
 You'll need to pass a GOV.UK Notify API key as an environment variable
 `NOTIFY_API_KEY`, and change the delivery method in [development.rb][]:
@@ -60,8 +60,9 @@ You'll need to pass a GOV.UK Notify API key as an environment variable
 config.action_mailer.delivery_method = :notify_email
 ```
 
-You'll also need to set a `GOVUK_NOTIFY_EMAIL_TEMPLATE_ID`, which might involve
-creating a template in Notify if your Notify service doesn't have one.
+You'll also need to set a `GOVUK_NOTIFY_EMAIL_TEMPLATE_ID` and
+`GOVUK_NOTIFY_SMS_TEMPLATE_ID`, which might involve creating a
+template in Notify if your Notify service doesn't have one.
 
 The template should have a Message of `((body))` only.
 
--- a/app/controllers/coronavirus_form/check_answers_controller.rb
+++ b/app/controllers/coronavirus_form/check_answers_controller.rb
@@ -48,7 +48,8 @@ private
 
   def smoke_tester?
     email = session_with_indifferent_access.dig(:contact_details, :email)
-    email.present? && email == Rails.application.config.courtesy_copy_email
+    mobile_number = session_with_indifferent_access.dig(:contact_details, :phone_number_texts)
+    (email.present? && email == Rails.application.config.courtesy_copy_email) || (mobile_number.present? && mobile_number == Rails.application.config.test_telephone_number)
   end
 
   def reference_number
--- a/app/controllers/coronavirus_form/check_answers_controller.rb
+++ b/app/controllers/coronavirus_form/check_answers_controller.rb
@@ -36,6 +36,7 @@ class CoronavirusForm::CheckAnswersController < ApplicationController
       )
 
       send_confirmation_email if session_with_indifferent_access.dig(:contact_details, :email).present?
+      send_confirmation_sms if session_with_indifferent_access.dig(:contact_details, :phone_number_texts).present? && mobile_number_provided?
     end
 
     reset_session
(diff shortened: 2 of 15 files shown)


README.md
```diff
@@ -51,17 +51,18 @@ to process the email queue.
...
 
 ```ruby
-config.action_mailer.delivery_method = :notify
+config.action_mailer.delivery_method = :notify_email
 ```
 
-You'll also need to set a `GOVUK_NOTIFY_TEMPLATE_ID`, which might involve
-creating a template in Notify if your Notify service doesn't have one.
+You'll also need to set a `GOVUK_NOTIFY_EMAIL_TEMPLATE_ID` and
+`GOVUK_NOTIFY_SMS_TEMPLATE_ID`, which might involvei creating a
```
bilbof:
```suggestion
`GOVUK_NOTIFY_SMS_TEMPLATE_ID`, which might involve creating a
```

app/controllers/coronavirus_form/check_answers_controller.rb
```diff
@@ -27,6 +27,7 @@ def submit
       )
 
       send_confirmation_email if session_with_indifferent_access.dig(:contact_details, :email).present?
+      send_confirmation_sms if session_with_indifferent_access.dig(:contact_details, :phone_number_texts).present? && mobile_number_provided?
     end
```
bilbof, line 40: I wonder if we also want to update the `smoke_tester?` method to check if the mobile number used is one used by the smoke tester?

app/controllers/coronavirus_form/check_answers_controller.rb
```diff
@@ -64,6 +74,14 @@ def user_email
...
   end
 
+  def user_mobile_number
+    TelephoneNumber.parse(session_with_indifferent_access.dig(:contact_details, :phone_number_texts), :gb).national_number(formatted: false)
+  end
+
+  def mobile_number_provided?
+    TelephoneNumber.valid?(session_with_indifferent_access.dig(:contact_details, :phone_number_texts), :gb, [:mobile])
```
bilbof, line 92: We'll need to remove whitespace from the mobile number here, if this isn't done elsewhere.

concourse/tasks/deploy-to-govuk-paas.yml
```diff
@@ -60,7 +61,8 @@ run:
       cf set-env govuk-coronavirus-vulnerable-people-form GA_VIEW_ID "$GA_VIEW_ID"
       cf set-env govuk-coronavirus-vulnerable-people-form METRICS_USERNAME "$METRICS_USERNAME"
       cf set-env govuk-coronavirus-vulnerable-people-form METRICS_PASSWORD "$METRICS_PASSWORD"
-      cf set-env govuk-coronavirus-vulnerable-people-form GOVUK_NOTIFY_TEMPLATE_ID "$GOVUK_NOTIFY_TEMPLATE_ID"
+      cf set-env govuk-coronavirus-vulnerable-people-form GOVUK_NOTIFY_EMAIL_TEMPLATE_ID "$GOVUK_NOTIFY_EMAIL_TEMPLATE_ID"
+      cf set-env govuk-coronavirus-vulnerable-people-form GOVUK_NOTIFY_SMS_TEMPLATE_ID "$GOVUK_NOTIFY_SMS_TEMPLATE_ID"
```
bilbof, line 65: `GOVUK_NOTIFY_TEMPLATE_ID` will still be set on the app, so we might want to unset that at some point once this is deployed.

config/locales/en.yml
```diff
@@ -191,6 +191,20 @@ en:
 
       <h2 class="govuk-heading-m">If you didn’t expect to see this page</h2>
       <p class="govuk-body">You may be seeing this page because you’ve turned off cookies in your browser. You’ll need to <a class="govuk-link" href="https://ico.org.uk/your-data-matters/online/cookies/" rel="external" target="_blank">turn cookies on</a> before you can use this service.</p>
+  sms:
+    confirmation:
+      body_text: |
+        Hello %{first_name} %{last_name} - we’ve received your registration.
```
bilbof: Do we want to limit the length of `first_name` or `last_name`, and also remove any special characters (e.g. `.`). Spammers could use this to send texts to people with e.g. some spam message, perhaps containing a link. This isn't specific to the SMS though..

spec/controllers/coronavirus_form/check_answers_controller_spec.rb
```diff
@@ -72,6 +72,29 @@
...
+        phone_number_texts: "07790 900000",
+      }
+
+      expect {
+        post :submit
+      }.to have_enqueued_mail(CoronavirusFormMailer, :confirmation_sms)
+        .with(a_hash_including(reference_number: "abc"), "07790900000").on_queue("mailers")
+    end
+
+    it "does not send a text message when no email address provided" do
```
bilbof:
```suggestion
    it "does not send a text message when no phone number for texting is provided" do
```

bilbof (commented):
Looks great, just a few comments.

It sounds like there are other things to consider (e.g. text content) too, so will approve when those are done.

app/controllers/coronavirus_form/check_answers_controller.rb
brucebolt: Good idea.  I've added a condition for this.

concourse/tasks/deploy-to-govuk-paas.yml
brucebolt: Cool, I'll do this manually after deployment of these changes

config/locales/en.yml
brucebolt: I'm hesitant to do this, since some names can be rather long and include special characters [1].  We have around 300 characters remaining in the SMS, so I don't think length is an issue.

Maybe we should have another story to perform some validation on these fields when the data is actually entered so anything that looks like a URL is disallowed, thereby preventing phishing attempts using this form?

[1]: https://www.bbc.co.uk/news/world-us-canada-52557291

app/controllers/coronavirus_form/check_answers_controller.rb
brucebolt: I think we will have done this when the number is stored in the contact details controller, since the number is being saved as `national_number` which will have already been formatted by the TelephoneNumber gem.

bilbof (approved):

brucebolt:
I've pushed an additional commit that will override the sender from "GOVUK" to the outgoing number for the National Shielding Service call centre.  Also configured this number in the Notify dashboard.

brucebolt:
This has been rebased against master and tested on staging.

I am awaiting confirmation on the sender name before we can merge and deploy to production.

brucebolt:
We've had confirmation the sender should be `GOVUK` (i.e. the default).  I have therefore removed the final commit for changing the sender.

config/locales/en.yml
```diff
@@ -192,6 +192,20 @@ en:
...
+
+        You don’t need to do anything else if you’ve had a letter from the NHS or your doctor confirming that you’re classed as clinically extremely vulnerable to coronavirus.
+
+        If you haven’t had a letter, contact your GP as soon as possible. Otherwise you may not get support.
+
+        We’ll check that you’re eligible, then you’ll start getting support if you asked for it. This may take time.
+
+        If you need urgent support and can’t rely on friends or neighbours, contact your local authority.
+
+        If your support needs change, go to https://www.gov.uk/coronavirus-extremely-vulnerable and answer the questions again.
```
bilbof: Do you know if this'll be one text based on the length? Just had a thought we might want a test for the length of this content, just in case.

bilbof (approved):
Excellent work 👍

config/locales/en.yml
brucebolt: It will cost multiple text messages to send (5x SMSs IIRC), but it will be delivered as a single message to the recipient's device

bilbof: Wow big. That's good to know 👍

PR #312, opened 2020-05-06
Fix phone number validation

bilbof:
Currently we don't accept some real phone numbers, including phone numbers in the format we use in our error message.

For some reason Github won't let me upload the screenshot of the error.

This change allows phone numbers to contain whitespace. The [TelephoneNumber](https://github.com/mobi/telephone_number) gem was rejecting numbers with whitespace. Additionally, some of our tests contained invalid phone numbers.

https://trello.com/c/JXUKXuf2/375-make-phone-number-field-more-permissive

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -247,7 +247,7 @@ en:
       postcode_format: "Enter a real postcode"
       negative_date: "Enter a real %{field}"
       date_not_a_number: "Enter %{field} as a number"
-      telephone_number_format: "Enter a telephone number, like 01632 960 001, 07700 900 982 or +44 0808 157 0192"
+      telephone_number_format: "Enter a telephone number, like 020 7946 0000, 07700900000 or +44 0808 157 0192"
     confirmation:
       title: Registration complete
       description: |
--- a/app/helpers/field_validation_helper.rb
+++ b/app/helpers/field_validation_helper.rb
@@ -104,7 +104,7 @@ module FieldValidationHelper
   end
 
   def validate_telephone_number(field, telephone_number)
-    if TelephoneNumber.parse(telephone_number, :gb).valid?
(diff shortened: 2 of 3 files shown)


issyl0:
Is it worth upstreaming a PR to the gem to enable it to accept and strip whitespace?

bilbof:
@issyl0 yeah probably, I'll do that. I'd like to get this merged in first, though.

issyl0 (approved):

config/locales/en.yml
```diff
@@ -247,7 +247,7 @@ en:
       postcode_format: "Enter a real postcode"
       negative_date: "Enter a real %{field}"
       date_not_a_number: "Enter %{field} as a number"
-      telephone_number_format: "Enter a telephone number, like 01632 960 001, 07700 900 982 or +44 0808 157 0192"
+      telephone_number_format: "Enter a telephone number, like 02033 445 566, 07711330382 or +44 0808 157 0192"
```
brucebolt: The choice of invalid telephone numbers here is deliberate, IIRC.  The numbers are from https://www.ofcom.org.uk/phones-telecoms-and-internet/information-for-industry/numbering/numbers-for-drama and are used so they definitely are not real number.

Details in the GOV.UK Design System: https://design-system.service.gov.uk/patterns/telephone-numbers/.

bilbof: That's interesting, thanks Bruce.

It looks like the gem or the underlying library is rejecting numbers from Ofcom's list, so I'll use Ofcom's numbers in the error text, but real (made up) numbers in the tests, if that's OK?

```ruby
TelephoneNumber.parse("07700900999", :gb).valid? # => false
```

As an aside, we should make sure that we don't send text messages in unit / smoke tests! :)

brucebolt: That sounds like a good plan :+1:

PR #322, opened 2020-05-13
ci: Remove step names and rename file to ci.yml

gclssvglx:
As per the [Actions conventions doc](alphagov/govuk-developer-docs#2533) once again.


issyl0 (approved):
⭐

PR #323, opened 2020-05-14
Add address lookup page

gclssvglx:
This completes the address lookup feature.

Adds a new page which handles looking up an addresses for a given postcode, selecting a specific address and confirming an address.

Also, stores a Unique Property Reference Number (UPRN) for the selected address.

We want to provide an address lookup to ensure that we collect correct addresses from users, and UPRN numbers where possible.

[Trello](https://trello.com/c/WoKUCfci/376-address-lookup-page)

Co-authored-by: Jon Hallam <jonathan.hallam@digital.cabinet-office.gov.uk>


bilbof:
Rebased 👍

app/helpers/address_helper.rb
```diff
@@ -9,7 +9,7 @@ def postcode_lookup(postcode)
     response = Faraday.get(postcode_url)
     details = JSON.parse(response.body)
 
-    raise AddressLookupError, details["error"] unless response.status == 200
+    raise AddressLookupError, response.status unless response.status == 200
 
     addresses = details["results"]
```
bilbof: `addresses` can be `nil`, so calling `addresses.map` below will fail. You could do `details.fetch("results", [])` and also check if `"totalresults"` is 0.

bilbof (commented):
There doesn't seem to be a way to edit a postcode after you enter an incorrect one. We should probably make that possible.

bilbof (dismissed):
We should the keys such as `POSTCODE_LOCATOR` (postcode) rather than parsing the raw `ADDRESS` field if possible.

E.g. if you enter the FCO canteen postcode "sw1a2ae":

Screenshot 2020-05-14 at 14 07 35

bilbof:
We should probably get some error messages approved by our content people.

Screenshot 2020-05-14 at 14 05 31

app/controllers/coronavirus_form/address_lookup_controller.rb
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
+
+class CoronavirusForm::AddressLookupController < ApplicationController
+  def submit
+    redirect_to check_your_answers_url if session[:check_answers_seen]
```
bilbof: It looks like this would prevent people from looking up a new address. Is that something we want to permit?

app/controllers/coronavirus_form/address_lookup_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+        flash.now[:validation] = [{ text: e.message, field: "postcode" }]
+        render controller_path, status: :unprocessable_entity
+      end
+    end
+  end
+
+private
+
+  def update_params
+    uprn_and_address = params[:address_string].split("|")
```
bilbof: If you use JSON in the input values you could then parse that as JSON here, rather than parsing a string.

E.g. 

```html
<select>
   <option value='{"uprn":"1234","postcode":"abc","address_line_1":"10 Downi..."..}'>
      10 Downing Street, London..
    </option>
</select>
```

```ruby
address = validate_address(JSON.parse(selected_address))
```

This should make it easier to parse and validate the selected address.

app/controllers/coronavirus_form/address_lookup_controller.rb
bilbof: Additionally, if someone goes back to this page, and re-enters a postcode, they get a 500 error, due to a DoubleRenderError (redirect_to and then a render controller_path called later).

app/controllers/coronavirus_form/address_lookup_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+
+    if params[:postcode].blank?
+      flash.now[:validation] = [{ text: I18n.t("coronavirus_form.questions.address_lookup.hint"), field: "postcode" }]
+      render controller_path, status: :unprocessable_entity
+    elsif params[:address_string].present?
+      update_params
+      redirect_to support_address_path
+    else
+      begin
+        results = helpers.postcode_lookup(params[:postcode])
```
bilbof: We should probably validate the postcode here using our postcode validator. No point making an API call with a bad postcode.

app/controllers/coronavirus_form/address_lookup_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  def generate_address_strings(results)
+    results.map do |address|
+      {
+        text: address.first.last,
+        value: "#{address.first.first} | #{address.first.last}",
+      }
+    end
+  end
```
bilbof, line 24: I think it'd be better if the API / helper class you have could provide the data in a format the controller requires, rather than putting extra logic into the controller. Ideally controllers are dumb and don't do anything other than call other classes

app/controllers/coronavirus_form/support_address_controller.rb
```diff
@@ -7,15 +7,25 @@ class CoronavirusForm::SupportAddressController < ApplicationController
...
+
     @form_responses = {
-      support_address: {
-        building_and_street_line_1: strip_tags(params[:building_and_street_line_1]&.strip).presence,
-        building_and_street_line_2: strip_tags(params[:building_and_street_line_2]&.strip).presence,
-        town_city: strip_tags(params[:town_city]&.strip).presence,
-        county: strip_tags(params[:county]&.strip).presence,
-        postcode: strip_tags(params[:postcode]&.gsub(/[[:space:]]+/, "")).presence,
-      },
+      support_address: helpers.compare_and_return(session[:support_address].symbolize_keys, given_address),
```
bilbof: I'm not really clear on the intent of the `compare_and_return` method.

app/errors/address_lookup_error.rb
```diff
@@ -1,5 +1,5 @@
 class AddressLookupError < StandardError
-  def initialize(message)
-    super(message)
+  def initialize(code)
+    super(I18n.t("coronavirus_form.api_errors.#{code}"))
```
bilbof: It would be worth having a default here, since it could be that you receive a different status code, or `nil` as the code.

config/locales/en.yml
```diff
@@ -233,6 +233,11 @@ en:
         Do not reply to this email - it’s an automatic message from an unmonitored account.
   coronavirus_form:
     submit_and_next: "Continue"
+    api_errors:
+      400: "Bad Request | E.g. missing query parameter, malformed syntax."
+      401: "Unauthorized | The client has not provided authentication or incorrect authentication."
+      404: "Not Found | The server has not found anything matching the Request-URI."
+      500: "Internal Server Error| Generic internal server error."
```
bilbof:
```suggestion
      500: "Internal Server Error | Generic internal server error."
```

However, I think a content person / service designer will likely want to write these.

spec/support/fill_in_the_form_steps.rb
```diff
@@ -65,6 +65,25 @@ def and_has_given_their_date_of_birth
...
+      expect(page.body).to have_content(I18n.t("coronavirus_form.questions.address_lookup.title"))
+      within find(".govuk-main-wrapper") do
+        address = "10, DOWNING STREET, LONDON, CITY OF WESTMINSTER, SW1A 2AA"
+        select(address, from: :address_string)
+        click_on I18n.t("coronavirus_form.submit_and_next")
+      end
+    end
+    VCR.turn_off!
+  end
+
```
bilbof, line 96: I think it would be useful to have tests for missing addresses and other errors here.

spec/controllers/coronavirus_form/address_lookup_controller_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+    context "valid postcode" do
+      it "renders the API results given a valid postcode" do
+        VCR.use_cassette "/address/200_individual_building" do
+          post :submit, params: { "postcode" => "SW1A2AA" }
+
+          expect(response).to have_http_status(:ok)
+          expect(response).to render_template(current_template)
+        end
+      end
+    end
```
bilbof, line 78: Would be good to have a test for what happens if they have already viewed the check_your_answers page here.

bilbof:
Great work so far 👍 I think there are a few changes that need to be made to the code, and to get designers / content folks to have a good look over this too :)

app/helpers/address_helper.rb
JonathanHallam: Heya, can you explain this a bit more. As far as we knew this was taking all of the addresses for a given postcode. Does this mean that there are postcodes without addresses? Do you have and example? 

We were expecting this to either return an empty array or to return a 400 before it got to this point.

JonathanHallam:
> There doesn't seem to be a way to edit a postcode after you enter an incorrect one. We should probably make that possible.

We think we're going to have to make another route in order to do this. Will make a start on that on Monday

bilbof:
Another thing that would be useful is if this could still work locally / in Heroku... I wonder if we could redirect people to the "enter address manually" page if - for whatever reason - the API isn't working (e.g. it goes down / the api key isn't working)

config/locales/en.yml
joelanman: https://design-system.service.gov.uk/patterns/service-unavailable-pages/

app/controllers/coronavirus_form/address_lookup_controller.rb
JonathanHallam: Now people can look up a new address but they have to go back to do it. They can either use the back links or browser buttons. Additionally the extra page we added seems to have fixed the 500 error

app/errors/address_lookup_error.rb
JonathanHallam: Have added in a default value :)

app/helpers/address_helper.rb
gclssvglx: Updated to use response.status.  Yes, there are postcodes without addresses (e.g. SW1A3AA) that we have accounted for.

app/controllers/coronavirus_form/address_lookup_controller.rb
gclssvglx: In order to make this work we had to make another API call to the uprn end-point.  We are not using JSON, but we are using all of the fields directly from the API response, so strings are no longer a problem.

app/controllers/coronavirus_form/address_lookup_controller.rb
gclssvglx: Done.

app/controllers/coronavirus_form/address_lookup_controller.rb
gclssvglx: Done.  We know use the fields from the API response directly.

app/controllers/coronavirus_form/support_address_controller.rb
gclssvglx: Fixed! See code.

config/locales/en.yml
gclssvglx: Have made changes to the copy and no longer use these responses from OS.

spec/support/fill_in_the_form_steps.rb
gclssvglx: Have added to other tests.

spec/controllers/coronavirus_form/address_lookup_controller_spec.rb
gclssvglx: Done.

app/controllers/coronavirus_form/address_lookup_controller.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+class CoronavirusForm::AddressLookupController < ApplicationController
+  def show
+    @addresses = helpers.postcode_lookup(session[:postcode])
+    render controller_path, status: :ok
+  rescue AddressLookupError, AddressNotFoundError => e
+    session[:flash] = e.message
+
+    if e.class.name == "AddressLookupError" && e.status == 401
```
bilbof: I would recommend raising a different error in the helper for the authentication error (e.g. `AddressAuthError`), then you can have a simpler controller, e.g.

```ruby
rescue AddressAuthError
      GovukError.notify("API Key is invalid!")
      redirect_to support_address_path
rescue AddressLookupError, AddressNotFoundError => e
      session[:flash] = e.message
      redirect_to postcode_lookup_url
```

app/controllers/coronavirus_form/address_lookup_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+      GovukError.notify("API Key is invalid!")
+      redirect_to support_address_path
+    else
+      redirect_to postcode_lookup_url
+    end
+  end
+
+  def submit
+    if session[:check_answers_seen]
+      redirect_to check_your_answers_url
```
bilbof: Do we want to save the new address if the user submit on this page again?

app/controllers/coronavirus_form/address_lookup_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+    else
+      address = helpers.uprn_lookup(params[:uprn])["results"][0]["LPI"]
```
bilbof: You could use `dig` here, I.e. `helpers.uprn_lookup(uprn).dig('results',0,'LPI')`, however, I wonder if this logic (including `convert_address(address)`) needs to happen in the controller? Could the helper just return the already converted address?

app/controllers/coronavirus_form/postcode_lookup_controller.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+class CoronavirusForm::PostcodeLookupController < ApplicationController
+  def show
+    if session[:flash].present?
```
bilbof, line 5: I wonder if you could avoid setting error messages and the session data by using the same controller for both the address picker and the postcode lookup? You'd have different views / endpoints, but you could store the errors etc in the controller instance variables and wouldn't need to do redirects if lookups failed, you'd just render a different view.

You wouldn't need to add and then remove values from the session, like `session[:postcode]`, since this could all be set in controller instance variables.

app/controllers/coronavirus_form/postcode_lookup_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+      session.delete(:flash)
+      @postcode_entered = true
+      render controller_path, status: :unprocessable_entity
+    else
+      render controller_path, status: :ok
+    end
+  end
+
+  def submit
+    redirect_to check_your_answers_url if session[:check_answers_seen]
```
bilbof: This'll prevent people from looking up another postcode, even if they enter a new one. Is that behaviour desired?

app/controllers/coronavirus_form/postcode_lookup_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+
+    if params[:postcode].blank?
+      flash.now[:validation] = [{ text: I18n.t("coronavirus_form.questions.address_lookup.hint"), field: "postcode" }]
+      render controller_path, status: :unprocessable_entity
```
bilbof, line 18: If I don't enter an address, the error message is "Select your address", when it should be "Enter your postcode". Is that intended?

app/controllers/coronavirus_form/postcode_lookup_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+    else
+      errors = helpers.validate_postcode("postcode", params[:postcode])
```
bilbof, line 20: I don't think you need to use `helpers` here, you can just call `validate_postcode`

bilbof: I wonder also whether `params[:postcode].blank?` could be done by the `validate_postcode` method? Then you could have less code in the controller (e.g. only one `render controller_path, status: :unprocessable_entity`), which is usually a good thing.

app/controllers/coronavirus_form/address_lookup_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+class CoronavirusForm::AddressLookupController < ApplicationController
+  def show
+    @addresses = helpers.postcode_lookup(session[:postcode])
+    render controller_path, status: :ok
+  rescue AddressLookupError, AddressNotFoundError => e
+    session[:flash] = e.message
+
+    if e.class.name == "AddressLookupError" && e.status == 401
+      GovukError.notify("API Key is invalid!")
+      redirect_to support_address_path
```
bilbof: It might be a bit confusing for users if this happens. Should we show them an error message?

app/controllers/coronavirus_form/support_address_controller.rb
```diff
@@ -1,53 +1,75 @@
...
+
       respond_to do |format|
         format.html { render controller_path, status: :unprocessable_entity }
       end
-    elsif session[:check_answers_seen]
-      session[:support_address] = @form_responses[:support_address]
-      redirect_to check_your_answers_url
     else
+      @form_responses = {
+        support_address: helpers.update_address(session[:support_address].symbolize_keys, given_address),
```
bilbof: Here could you set `@form_responses = helpers.update_address ...` rather than it being a hash?

app/errors/address_lookup_error.rb
```diff
@@ -1,5 +1,8 @@
 class AddressLookupError < StandardError
-  def initialize(message)
-    super(message)
+  attr_reader :status
+  def initialize(status)
+    @status = status
+    @status = "error" unless I18n.t("coronavirus_form.api_errors").keys.include?(@status)
+    super(I18n.t("coronavirus_form.api_errors.#{@status}"))
```
bilbof: You could use a `default` here instead of the `unless` line: https://guides.rubyonrails.org/i18n.html#defaults

e.g. `I18n.t("coronavirus_form.api_errors.#{@status}", default: "coronavirus_form.api_errors.error")`

app/helpers/address_helper.rb
```diff
@@ -1,19 +1,104 @@
 # frozen_string_literal: true
 
 module AddressHelper
-  API_URL = "https://api.ordnancesurvey.co.uk/places/v1/addresses/postcode?dataset=LPI"
+  UPRN = %w[UPRN].freeze
+  SAO = %w[ORGANISATION SAO_START_NUMBER SAO_START_SUFFIX SAO_END_NUMBER SAO_END_SUFFIX SAO_TEXT].freeze
+  PAO = %w[PAO_START_NUMBER PAO_START_SUFFIX PAO_END_NUMBER PAO_END_SUFFIX PAO_TEXT].freeze
```
bilbof, line 6: Oh my word, this looks hard! You must know so much about addresses now 👏

bilbof (commented):
Great progress 👍 I've reviewed the controllers / logic, but wanted to give some immediate feedback, before re-reviewing.

I found the postcode lookup request quite slow (maybe a few seconds). Is there a way to do just one request to the OS API? I wasn't sure why we'd need to make two requests there.

JonathanHallam:
> Great progress 👍 I've reviewed the controllers / logic, but wanted to give some immediate feedback, before re-reviewing.
> 
> I found the postcode lookup request quite slow (maybe a few seconds). Is there a way to do just one request to the OS API? I wasn't sure why we'd need to make two requests there.

@bilbof 

So we could do it with one request but that would mean that we'd either have to rely heavily on JS or be saving a potentially massive hash to the session. The problem is that the user has to choose their address from the list. This kind of makes it necessary to have an extra route which, along with wanting to use the json fields to put chosen address data into the support-address page is what necessitates the second call. 

It's also worth noting that the initial call to the api after you input a postcode would take exactly the same amount of time. The extra time spent by the 2nd call is only what's spent between clicking continue after choosing an address from the dropdown and seeing that address on the support-address page.

app/controllers/coronavirus_form/address_lookup_controller.rb
JonathanHallam: We do. We figured that if OS was returning a 401 it meant that someone had revoked our key and that there was nothing the user could do. In the event that this happens it redirects to a blank support-address form with a message that says `An error has occurred, please enter your address manually`

You can test this locally by replacing your api key with nonsense and looking up an address.

app/helpers/address_helper.rb
JonathanHallam: yeah, more than we'd like! If you want talking through this then we'd be happy to. Also happy to do a brief show and tell to the other devs on how addresses work once this story's in if you think that'd be useful :)

app/controllers/coronavirus_form/address_lookup_controller.rb
JonathanHallam: Nice shout, have just pushed this change up :)

app/controllers/coronavirus_form/address_lookup_controller.rb
JonathanHallam: Nice spot on the dig, we've adjusted. On the 2nd point, we'd rather leave these as 2 separate calls. This makes it easier to test because it avoids having to chain methods together in the helper. We could maybe avoid this by having a third method along the lines of `lookup_address_by_uprn_and_convert` but I really don't like that.

app/controllers/coronavirus_form/postcode_lookup_controller.rb
JonathanHallam: When you first reviewed this we did have everything in one controller but that was mainly because there was only one route. In order to improve the format in which full addresses we displayed after selection we felt that we needed the extra route. Going back to a single controller way of doing things would help solve the issues that you mention above but would make that controller quite heavy with logic which we'd be keen to avoid. 

You'd also still be in the position where you'd have to redirect to the support address path in the event you get a 401 out of the OS API. We think it's best to show that error message on the support-address page so even if we were to do this at the very least we'd need to be storing that error message in the session.

Storing the error message in the session is a bit of an anti pattern but having the two controllers does cut down on controller logic and avoid us having a controller that was several hundred lines long. Happy to talk about this offline.

app/controllers/coronavirus_form/postcode_lookup_controller.rb
JonathanHallam: Ooops, absolutely not. Have fixed, good spot 🥇

app/helpers/address_helper.rb
bilbof: Definitely that would be useful :) I'm sure this would be something for the wider gov.uk show and tell, too.

app/controllers/coronavirus_form/postcode_lookup_controller.rb
JonathanHallam: We tried removing the helpers calls here and it caused a few method errors in the tests.

Just had a look at this and yes it does seem a bit unnecessary to have that call here but the `validate_postcode` method already does a check for a blank postcode. It does this to stop itself falling over when it gets to the regex that checks format. It is possible to input an error if postcode is blank in the `validate_postcode` method. If we want to move this check to the `validate_postcode` method we have two options. 

One is to remove `POSTCODE` from the `REQUIRED_FIELDS` constant in the `support_address_controller`. This will stop the validation happening twice but is quite horrible because we'd kind of be saying that postcode both was and wasn't mandatory. 

The other would be to move the `validate_missing_fields` method from the `support_address_controller` into the `field_validation_helper`. This is doable but would mean moving quite a bit of stuff around. It would also make the i18n structure a bit brittle because we'd want it to be universal for all of the required fields across the form. 

I think this would be worth more thought if we started doing this in more cases. As it stands though, postcode is a bit of a snowflake because of the regex validation that we're doing in the `field_validation_helper`. Because of this I don't think it's worth the downsides / effort.

app/controllers/coronavirus_form/support_address_controller.rb
JonathanHallam: nice, this clears it up some, thanks :)

app/errors/address_lookup_error.rb
JonathanHallam: didn't know this was a thing, nice!

app/controllers/coronavirus_form/postcode_lookup_controller.rb
gclssvglx: Removed the redirect from here as it was possible to enter a new address without saving it ini the session.

app/controllers/coronavirus_form/address_lookup_controller.rb
gclssvglx: See comment on `postcode_lookup_controller`.

bilbof (commented):
I'm happy for this to be merged. Great work @JonathanHallam @gclssvglx! 👏 

It'd be good to have the product and design folks on the team give this a thorough review / QA with the Heroku app, so if you could share the running review app with them that'd be a good next step. I'll add a green tick here once they've approved.

I'd quite like some brief docs on how this works, not only so we can maintain this repo, but also so others can easily copy this pattern elsewhere. If you have ideas of how this could be put into a pattern/component, those docs would be a good place to put them.

bilbof (approved):
🎉

joelanman (changes requested):
The current postcode validation on the service strips out any whitespace before validating - I think we should stick with that logic, as we know what postcode the user meant, and whitespace is hard for a user to 'fix'.

The address lookup pattern in the Design System plays back the postcode on the dropdown page so a user can check it, and has a link to change it (back to the previous page) if its not correct - it also has a shorter label for the dropdown, can we have 'Select your address'?

joelanman:
Sorry forgot one - if the user enters an unrecognised postcode, they just get a validation error and cannot continue. We should probably link to 'Enter your address manually' in case of some issue that means the lookup isn't working in 100% of cases

app/views/coronavirus_form/address_lookup.html.erb
```diff
@@ -0,0 +1,39 @@
...
+<p class="govuk-body">
+  <%= link_to t("coronavirus_form.questions.address_lookup.cant_find_address.label"), support_address_path, class: "govuk-link" %>
+</p>
+
+<%= render "govuk_publishing_components/components/button", text: t("coronavirus_form.submit_and_next"), margin_bottom: true %>
+<% end %>
+
+<p class="govuk-body">
+  <%= t("coronavirus_form.questions.address_lookup.copyright") %>
+</p>
```
joelanman: can we remove this? Not sure why it's been added

joelanman (changes requested):

app/views/coronavirus_form/address_lookup.html.erb
JonathanHallam: This was on the AC for the card. I'm guessing that it's required because we're using copyrightable data from Ordinance Survey.

StephenGill: Do we have a link to the licence?

joelanman: it was on the card, but no one seems to know why. It's unusual to have a line like this in a service

gclssvglx: Removed.

joelanman (approved):
Great work!

bilbof:
LGTM. If we have stakeholder approval lets merge this tomorrow morning :)

joelanman:
Sorry, looked into the copyright thing - looks like the correct line is:

>Contains Royal Mail data © Royal Mail copyright and database rights 2020

Can we make it use the small font, and space it away from the form - mockup:

image

joelanman:
Looks great thanks

PR #328, opened 2020-05-14
Make NHS Number compulsory

gclssvglx:
What
----

We want to make NHS number compulsory because people cannot receive help without an NHS number and quite a number of people are skipping this step.

We also want to update the existing `What is your NHS number?` page with additional supporting, add more help content and reflect the desired design.

Ensure the results page is still ok.

Screen shots
-------------

### Update NHS page design

screencapture-localhost-5000-nhs-number-2020-05-14-12_28_08

### NHS Number should be mandatory

screencapture-localhost-5000-nhs-number-2020-05-14-14_14_22

### Results page

screencapture-localhost-5000-check-your-answers-2020-05-14-13_54_36

Links
-----

[Trello card](https://trello.com/c/7pp5RLZM/444-make-nhs-number-compulsory)


brucebolt (approved):
Great work 👍.

I'd suggest checking with the data team before merging, just to make sure they are not using the `know_nhs_number` field for anything.

gclssvglx:
> Great work 👍.
> 
> I'd suggest checking with the data team before merging, just to make sure they are not using the `know_nhs_number` field for anything.

Thanks @brucebolt - yeah, I've put a message on the covid-engineering-team slack channel - waiting for a reply :-)

app/views/coronavirus_form/nhs_number.html.erb
```diff
@@ -15,20 +15,35 @@
       "novalidate": "true"
     ) do %>
 
+    <%= render "govuk_publishing_components/components/title", {
+      title: t("coronavirus_form.questions.nhs_number.title"),
+      margin_top: 0,
+      margin_bottom: 3,
+    } %>
+
+    <%= render "govuk_publishing_components/components/hint", {
```
alex-ju: the hint component is not meant to be used in solitude, but part of a fieldset/input field to help describing inputs. I would recommend outputting this as plain text `<%= t("coronavirus_form.questions.nhs_number.hint") %>` and wrap [each content line in the yml](https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/pull/328/files#diff-37c9b835927e3c363588f172de9bb6faR362-R365) in a `<p class="govuk-body">` – aside from this suggestion, frontend looks good 👍

JonathanHallam (approved):
I'm happy with this pending the changes that Alex has asked for to the frontend :)

app/views/coronavirus_form/nhs_number.html.erb
gclssvglx: Thanks @alex-ju - have updated as requested.

alex-ju (approved):
💯

config/routes.rb
```diff
@@ -43,11 +43,7 @@
     get "/contact-details", to: "contact_details#show"
     post "/contact-details", to: "contact_details#submit"
 
-    # Question 8.0: Do you know your NHS number?
-    get "/know-nhs-number", to: "know_nhs_number#show"
-    post "/know-nhs-number", to: "know_nhs_number#submit"
```
bilbof, line 48: Annoyingly, removing the routes to the page will break the form for people already on the `/know-nhs-numbers` page and the previous page.

E.g. I am a user on the `/know-nhs-number` page during a deploy. I submit the form during or after the deploy, which will result in a 404 error.

I wonder if we could solve that for this one, by adding a redirect to the NHS number page for both of these routes, then remove them later, in an A AB B style deploy?

bilbof (commented):
I think you'll also need to update the schema for this.

app/controllers/coronavirus_form/nhs_number_controller.rb
```diff
@@ -18,9 +18,11 @@ def submit
         format.html { render controller_path, status: :unprocessable_entity }
       end
     elsif session[:check_answers_seen]
+      session[:know_nhs_number] = I18n.t("coronavirus_form.questions.know_nhs_number.options.option_yes.label")
```
bilbof: Move this up the block so you only have to have this line in there once.

Thanks for adding a test for this!

gclssvglx: Fixed - thanks @bilbof

config/routes.rb
gclssvglx: Thanks @bilbof - nice catch thanks!  I've tested these manually, and as they are temporary - I guess we won't need an automated test for these.  Shout if we do though.

bilbof (approved):
Great stuff 👏

PR #329, opened 2020-05-14
Add json schema

leenagupte:
Trello: https://trello.com/c/lGhAtvlg

# What's changed?
Creates a JSON schema for the vulnerable people service, similar to that added to the business form: https://github.com/alphagov/govuk-coronavirus-business-volunteer-form/pull/270.

Uses the schema in tests, and report errors to Notify at runtime if user submissions aren't valid against the schema.

Later we'll use this schema in an ActiveRecord validation, when we're sure user submissions are valid against the schema (otherwise we may prevent people from submitting their data).

# Why?
This should make life easier for the data team; we'll be able to ensure that what we provide them is valid and will result in the person submitting the form receiving help.


bilbof (approved):
Excellent work 👍

PR #335, opened 2020-05-15
Accept `session_id` in the schema

leenagupte:
Follows on from: https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/pull/329

It looks like `session_id` is being stored in the session, and therefore propagated to the DB.

Should fix: https://sentry.io/organizations/govuk/issues/1667664108/?project=5170680&referrer=slack

### DIFF
--- a/config/schemas/form_response.json
+++ b/config/schemas/form_response.json
@@ -161,6 +161,9 @@
     },
     "_csrf_token": {
       "type": "string"
+    },
+    "session_id": {
+      "type": "string"
     }
   },
   "definitions": {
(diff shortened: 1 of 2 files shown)


bilbof (commented):
👍 nice one. I'd prefer instead if we could remove the `session_id` from the submission, since the data team don't need this ID.

leenagupte:
> 👍 nice one. I'd prefer instead if we could remove the `session_id` from the submission, since the data team don't need this ID.

So try and save `FormResponse: session.except(:session_id)` ?
As we're just dumping the `session` wholesale into `FormResponse` we're passing quite a few things that the data team don't need, for example:

- check_answers_seen
- _csrf_token
- current_path
- previous_path

There are probably others. I think it's worth doing a proper review of the data we're storing.

bilbof:
@leenagupte that's a good idea. I've written up a story for that https://trello.com/c/SsnvWiaA/175.

bilbof (approved):

PR #336, opened 2020-05-15
Task/add sidekick information to prometheus

danwilliams:
### Changes

The requirements around what metrics should be provided, and what they were to be used for was slightly ambiguous, so this PR endeavours to match the requirements to what is provided by `Sidekiq::Stats`. What follows is a description.

**Required metric:** Time to send an email

**Provided:** [Queue latency](https://www.rubydoc.info/github/mperham/sidekiq/Sidekiq%2FQueue:latency) for the email queue (**prometheus metric name:** `sidekiq_mailers_queue_latency`). This is a rough indicator of the time it will take to send a given email from the moment it is queued. 

**Required metric:** Email job failure rate

**Provided:** Job failure rate for sidekiq as a whole (**prometheus metric name:** `sidekiq_failed_jobs`) and in addition a count of the dead sidekiq jobs (**prometheus metric name:** `sidekiq_dead_jobs`). The failure rate on a per job/queue basis is not data that is collected by sidekiq and is thus not available in the `Sidekiq::Stats` API. 

N.B: Currently the only job ran by the `govuk-coronavirus-vulnerable-people-form` service is the email job, so, for now, the two are equivalent.

**Required metric:** Email job volume

**Provided metric:** The volume of jobs in the email job queue (**prometheus metric name:** `sidekiq_mailers_queue_size`). Sidekiq doesn't keep records of the jobs that have been completed on a per queue basis, so historical information isn't available here. The same point as above re emails currently being the only job on the service holds true, and we provide the metric `sidekiq_processed_jobs` that tracks the total number of processed jobs.

### How to review

The metrics test has been extended to check that some sidekiq information is included, tests can be run the usual way via: 

`rake exec bundle`

In addition one can see the individual metrics by running:

`foreman start`  and going to `http://localhost:5000/metrics`

One can also configure a prometheus job to scrape the metrics endpoint, and observe that prometheus is picking up the metrics.

## Links

https://trello.com/c/QzchHd65/265-add-monitoring-for-sidekiq

### DIFF
--- a/lib/sidekiq_stats_prometheus_exporter.rb
+++ b/lib/sidekiq_stats_prometheus_exporter.rb
@@ -1,39 +1,37 @@
 require "sidekiq/api"
 require "prometheus/middleware/exporter"
 
-module Services
-  class SidekiqStatsPrometheusExporter < Prometheus::Middleware::Exporter
-    def initialize(app, options = {})
-      super(app, options)
+class SidekiqStatsPrometheusExporter < Prometheus::Middleware::Exporter
+  def initialize(app, options = {})
+    super(app, options)
 
-      @prometheus = Prometheus::Client.registry
+    @prometheus = Prometheus::Client.registry
 
-      @sidekiq_total_processed_jobs = @prometheus.gauge(:sidekiq_processed_jobs, docstring: "The number of jobs processed by sidekiq")
-      @sidekiq_total_failed_jobs = @prometheus.gauge(:sidekiq_failed_jobs, docstring: "The number of failed sidekick jobs")
-      @sidekiq_total_dead_jobs = @prometheus.gauge(:sidekiq_dead_jobs, docstring: "The number of jobs dead sidekiq jobs")
-      @sidekiq_mailers_queue_size = @prometheus.gauge(:sidekiq_mailers_queue_size, docstring: "The number of jobs in the 'mailers' queue")
-      @sidekiq_mailers_queue_latency = @prometheus.gauge(:sidekiq_mailers_queue_latency, docstring: "The latency of the sideqkiq 'mailers' queue")
-    end
+    @sidekiq_total_processed_jobs = @prometheus.gauge(:sidekiq_processed_jobs, docstring: "The number of jobs processed by sidekiq")
+    @sidekiq_total_failed_jobs = @prometheus.gauge(:sidekiq_failed_jobs, docstring: "The number of failed sidekick jobs")
+    @sidekiq_total_dead_jobs = @prometheus.gauge(:sidekiq_dead_jobs, docstring: "The number of jobs dead sidekiq jobs")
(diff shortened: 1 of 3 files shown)


lib/sidekiq_stats_prometheus_exporter.rb
```diff
@@ -0,0 +1,37 @@
...
+  def gather_sidekick_metrics
+    stats = Sidekiq::Stats.new
+    mailer_queue = Sidekiq::Queue.new("mailers")
+    @sidekiq_total_processed_jobs.set(stats.processed)
+    @sidekiq_total_failed_jobs.set(stats.failed)
+    @sidekiq_total_dead_jobs.set(stats.dead_size)
+    @sidekiq_mailers_queue_size.set(mailer_queue.size)
+    @sidekiq_mailers_queue_latency.set(mailer_queue.latency)
+  rescue StandardError => e
+    Rails.logger.error ["Error gathering sidekiq stats", e.message, *e.backtrace].join($INPUT_RECORD_SEPARATOR)
```
bilbof, line 26: It would be great for us to hear about this kind of error.

With rescued errors we usually send the error to Sentry too, so we get notified: https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/blob/master/app/controllers/coronavirus_form/check_answers_controller.rb#L25

It looks like if there is an error here those stats would not appear in the results, is that the desired behaviour, rather than the endpoint returning a 500?

lib/sidekiq_stats_prometheus_exporter.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  def respond_with(format)
+    gather_sidekick_metrics
+    [
+      200,
+      { "Content-Type" => format::CONTENT_TYPE },
+      [format.marshal(@registry)],
+    ]
+  end
+end
```
bilbof, line 37: Great job here.

We could get extra stats by using gauges / counters within the jobs themselves, e.g. https://github.com/prometheus/client_ruby#metrics. Do you think this could fill in any gaps that using `Sidekiq::Stats` alone would leave?

bilbof (approved):
Looks good. I had a couple questions, but feel free to merge.

danwilliams:
Summarising discussion elsewhere:

As sidekiq is a separate process, and prometheus metrics are not automatically forwarded to the rails process, gathering custom metrics from jobs necessitates either:

a) Finding a way to relay information from sidekiq to the rails process, which then inserts the metrics into the rails processes metrics. This could take the form of using redis as a pub-sub bus, or perhaps querying the sidekiq process on demand for metrics gathered by something like https://github.com/fastly/sidekiq-prometheus, assuming we know where sidekiq lives.

b) Using something  off the shelf (like the above) and have prometheus query sidekiq directly. Note - this option wasn't liked so much last time it was suggested as it would add complexity to the prometheus setup.

PR #338, opened 2020-05-15
Use `scss_lint-govuk` gem rather than `scss-lint`

issyl0:
What
----

- We should use our wrappers, as that's got GOV.UK-preferred config.
- Spotted this because of a Dependabot PR to bump `scss-lint` the
  upstream gem.

How to review
-------------

Observe that CI passes.

Links
-----

https://trello.com/c/0WEIi10A/450-use-scsslint-govuk-gem-rather-than-scss-lint

### DIFF
--- a/.scss-lint.yml
+++ b/.scss-lint.yml
@@ -1,0 +1,1 @@
+plugin_gems: ["scss_lint-govuk"]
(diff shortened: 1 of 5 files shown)


alex-ju (approved):
💯

PR #345, opened 2020-05-18
Spike using mailcheck gem to get email address suggestions

leenagupte:
Trello: https://trello.com/c/Y7Yyko42

# What?

Spike a way to get suggestions to common email typos.

# Why?

As of 10th May, we have sent 83k confirmation emails to registrants on the vulnerable people service. Of these, almost 1,500 have failed due to incorrect email addresses.

Doing a very unscientific analysis of a few hundred failures, there seem to be some common typos:
- .con instead of .com
- .cpm instead of .com
- .cim instead of .com
- .cm instead of .com
- .couk instead of .co.uk
- hotail.co.uk instead of hotmail.co.uk
- fmail.com instead of gmail.com

Additionally, we have seen over 2.5k "invalid email address" validation responses from Notify through Sentry: https://sentry.io/organizations/govuk/issues/1646807753/?project=5170680&query=is%3Aunresolved

A very common issue in the Sentry errors is users entering two dots before com, possibly because they are using a .com button on their phone's keyboard.

In total, this means 3,600 people haven't received a confirmation email because they entered an invalid or incorrect email address.

One suggestion from the Design System is checking for common misspelling and warning the user.

# What's next?

If we decide to take this approach we need to do some work to determine what this "suggestion" looks for the user. How do we let users proceed if they know that their email address is in fact correct?


bilbof (commented):
Looks good!

pixeltrix:
I would recommend this approach over the one in #350, however there are issues with the mailcheck gem which I evaluated for the Petitions website. In general, the domains are US biased and so you may not get a very good hit ratio on your data set plus the approach of using string distance to suggest an autocorrect can result in false positives, e.g. mail.com is only one letter away from gmail.com.

In the end I created a corpus of email addresses that had failed to validate their signatures and then [created rules to autocorrect them][1] where there were obvious typos. Since in our signature flow we re-present the email address as a way of confirming it we can apply the autocorrect in the first step and let the user correct the email address if it's been applied incorrectly in the second step - unsure whether that fits in with your flow, but it's relatively straightforward to adapt the pattern to suggest corrections.

FYI, one pattern we saw that you don't have listed above is the use of .co.uk on a domain that should be .com, e.g. gmail.co.uk should be gmail.com.
  
[1]: https://github.com/alphagov/e-petitions/commit/2225ed2287585a090e97956268eb2ad131529457#diff-5d30de8255fead54ce968b6b279ddbc5

spec/helpers/field_validation_helper_spec.rb
```diff
@@ -127,4 +127,17 @@
       expect(invalid_fields).to eq [{ field: "email_address", text: I18n.t("coronavirus_form.errors.email_format") }]
     end
   end
+
+  describe "#email_address_suggestion" do
+    let(:valid_email_address) { "me@example.com" }
+    let(:invalid_email_address) { "me@example..com" }
```
bilbof, line 133: Would be good get a few more examples in here e.g.

```ruby
    %w(
      govuk-coronavirus-services@digital.cabinet_office,gov.uk
      govuk-coronavirus-services@digital
      govuk-coronavirus-services@
      @gov.uk
      me@example..com
      a@b
    ).each do |invalid_email|
      it "does not move to next step with an invalid email address of #{invalid_email}" do
        expect(email_address_suggestion(invalid_email_address)).to eq(valid_email_address)
      end
    end
```

bilbof (approved):

spec/helpers/field_validation_helper_spec.rb
leenagupte: Isn't that just testing the gem though?

leenagupte:
Implemented in: https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/pull/368

PR #349, opened 2020-05-19
Update SMS content

brucebolt:
What
----

The confirmation SMS was one character too long (limit is 918 characters) which resulted in the message not sending when deploying to staging.

Therefore making two changes:
- Content shortened for confirmation SMS to be less than 918 characters
- Added a test for this (and all future) SMS content, to ensure we catch this before deploying

How to review
-------------

Review code and CI tests pass.

Links
-----

Trello card: https://trello.com/c/8L0b1iIg


bilbof (approved):
👍

PR #350, opened 2020-05-19
Spike truemail

leenagupte:
Trello: https://trello.com/c/Y7Yyko42

# What?

Spike a way to catch most invalid email addresses by checking that the email exchange exists.

# Why?

Just using a regex doesn't seem to work as it lets some slip through the net.

As of 10th May, we have sent 83k confirmation emails to registrants on the vulnerable people service. Of these, almost 1,500 have failed due to incorrect email addresses.

Doing a very unscientific analysis of a few hundred failures, there seem to be some common typos:
- .con instead of .com
- .cpm instead of .com
- .cim instead of .com
- .cm instead of .com
- .couk instead of .co.uk
- hotail.co.uk instead of hotmail.co.uk
- fmail.com instead of gmail.com

Additionally, we have seen over 2.5k "invalid email address" validation responses from Notify through Sentry: https://sentry.io/organizations/govuk/issues/1646807753/?project=5170680&query=is%3Aunresolved

A very common issue in the Sentry errors is users entering two dots before com, possibly because they are using a .com button on their phone's keyboard.

In total, this means 3,600 people haven't received a confirmation email because they entered an invalid or incorrect email address.

One suggestion from the Design System is checking for common misspelling and warning the user.

# What's next?

If we decide to take this approach we need to do some work to determine what level of validation we want to use.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -22,6 +22,7 @@ gem "sentry-raven", "~> 3.0"
 gem "sidekiq", "~> 6.0"
 gem "telephone_number", "~> 1.4"
 gem "timecop"
+gem "truemail", "~> 1.7.1"
 gem "uglifier", "~> 4.2"
 
 gem "prometheus-client", "~> 2.0"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -349,6 +349,8 @@ GEM
       docile (~> 1.1)
       simplecov-html (~> 0.11)
     simplecov-html (0.12.2)
+    simpleidn (0.1.1)
+      unf (~> 0.1.4)
     sprockets (3.7.2)
       concurrent-ruby (~> 1.0)
       rack (> 1, < 3)
@@ -362,6 +364,8 @@ GEM
     thread_safe (0.3.6)
     tilt (2.0.10)
     timecop (0.9.1)
+    truemail (1.7.1)
+      simpleidn (~> 0.1.1)
     tzinfo (1.2.7)
       thread_safe (~> 0.1)
     uglifier (4.2.0)
@@ -427,6 +431,7 @@ DEPENDENCIES
   simplecov (~> 0.16)
   telephone_number (~> 1.4)
   timecop
+  truemail (~> 1.7.1)
   uglifier (~> 4.2)
   vcr
   webdrivers
--- a/app/helpers/field_validation_helper.rb
+++ b/app/helpers/field_validation_helper.rb
@@ -86,7 +86,7 @@ module FieldValidationHelper
   end
 
   def validate_email_address(field, email_address)
-    if email_address.match?(URI::MailTo::EMAIL_REGEXP)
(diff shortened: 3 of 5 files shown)


bilbof (commented):
This is a cool approach 👍

I wonder if we could try [Truemail](https://github.com/rubygarage/truemail) without using it for validations, to see what it brings up, since I'm not sure whether it'll result in real email addresses getting rejected.

The gem also doesn't look like it is actively maintained, so I wonder if there are alternatives for doing the DNS record / email server requests?

I wonder what the effect of these requests would be on latency. Are they fast?

leenagupte:
@bilbof 
> The gem also doesn't look like it is actively maintained, so I wonder if there are alternatives for doing the DNS record / email server requests?
The gem was last updated 9 days ago, so I think we're ok on the maintenance front 😄 

I've tried calling `valid?` on a few of the errors reported by Sentry, and it's caught them correctly

> I wonder if we could try [Truemail](https://github.com/rubygarage/truemail) without using it for validations, to see what it brings up, since I'm not sure whether it'll result in real email addresses getting rejected.

How do you want to try it? 

> I wonder what the effect of these requests would be on latency. Are they fast?

It's fast locally

The reason I like Truemail is that you can tune the config to your preferences.

bilbof:
> I've tried calling valid? on a few of the errors reported by Sentry, and it's caught them correctly

Nice! I'm worried since there are so many email addresses, it might prevent valid emails going through.

> How do you want to try it?

Some ideas of how to try it in order of risk (in all cases we could also send an error to Sentry with details):

- We could alert us and not tell the user about the issue
- We could warn the user, but not prevent them from progressing (some flag that gets set on the first try, or something)
- We could validate it for real, and not allow the user to progress

Do you think the gem will run OK in production? I'm not sure what happens if we send lots of requests to gmail.com asking if emails exists, or if the gem has good error handling if those requests die, and so on.

pixeltrix:
I would recommend not using DNS and SMTP checks for email validation as they are very resource intensive to do in a web request and may take many seconds to respond. Whilst the truemail gem does at least use a thread-aware DNS resolver, it doesn't prevent bad DNS resolution from blocking the thread and the SMTP connection may take many seconds to respond.

Even then they're not really that useful since many common typos have been registered and have MX records and even listening SMTP servers - for example hotmial.com is almost certainly a typo but would pass a DNS MX check and SMTP connection test. Sometimes such registrations are bad actors but others such as common typos for `@btinternet.com` addresses are registered to redirect mail to the intended recipient.

There is also an added security risk in that either someone could flood the DNS server your using for resolution and block access to the website or they could set up a slow DNS server and/or SMTP server and flood the website with checks to those email addresses and again block access to the website.

leenagupte:
We've decided to implement email suggestions using the mailcheck gem instead. See: https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/pull/368

PR #356, opened 2020-05-20
Remove unnecessary keys from the FormResponse

steveholmes:
What
----

We're submitting a few fields in `FormResponse` records that shouldn't exist there. These keys, such as `_csrf_token` are being sent through to the data team.

The changes here are to sanitise the session data and remove unnecessary keys when persisting the form response to the DB.

How to review
-------------

After completing the questions check that the response data in the DB does not contain any of 
```
check_answers_seen
_csrf_token
current_path
previous_path
session_id
```
Links
-----

https://trello.com/c/SsnvWiaA/463-remove-unnecessary-fields-from-formresponse

### DIFF
--- a/app/controllers/coronavirus_form/check_answers_controller.rb
+++ b/app/controllers/coronavirus_form/check_answers_controller.rb
@@ -94,7 +94,7 @@ private
   end
 
   def sanitised_session
-    session_with_indifferent_access.except(:session_id, :_csrf_token, :current_path, :previous_path, :check_answers_seen)
+    session_with_indifferent_access.except(:_csrf_token, :current_path, :previous_path, :check_answers_seen)
   end
 
   def contact_gp?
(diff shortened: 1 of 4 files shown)


issyl0:
Thanks for your work here!

Sorry to be super pedantic but can we have some of the PR description (the what and the why) in the commit message please? As it's a single commit this should be fairly straight forward. You've linked to the Trello card, which is great, but Trello boards aren't necessarily long-lived and future us (or whoever comes in next) might need to go back to the individual commit for context.

bilbof (approved):
Looks good, just needs a rebase in order to merge.

benvand:
@steveholmes We need to coordinate this release with https://github.com/alphagov/covid-engineering/pull/285
I also think the reporting team is relying on session_id

steveholmes:
Now confirmed that session_id is not in fact used, so will remove session_id

benvand:
@steveholmes we're ready for these fields to disappear now! 🎉

PR #358, opened 2020-05-21
Only show food related questions if someone has food needs

brucebolt:
What
----

- Changes the order of questions, so the "do you have any special dietary requirements" and "can you carry a box inside" questions come immediately after the "do you have a way of getting essential supplies delivered at the moment" question.
- Only shows the two later food questions if the user does not have a way of getting food.
- Updates the "previous" link target for subsequent food pages based on the user's answers.
- Removes the skippable questions from the "check your answers" page when the user has not answered them.
- Ensures a user is shown the skippable questions if they change their answer to say they do have problems getting food.

Not to be merged until approved by DWP call centre, and the data processing team.

How to review
-------------

- Review code changes.
- Run app locally, or use preview, to examine behaviour of `/essential-supplies` and subsequent questions when the different responses are given.  Also review the behaviour of changing your answer to the food questions.

Links
-----

Trello card: https://trello.com/c/IWPGeYTp

### DIFF
--- a/app/controllers/coronavirus_form/basic_care_needs_controller.rb
+++ b/app/controllers/coronavirus_form/basic_care_needs_controller.rb
@@ -27,6 +27,8 @@ class CoronavirusForm::BasicCareNeedsController < ApplicationController
 private
 
   def previous_path
-    carry_supplies_path
+    return carry_supplies_path if @form_responses[:essential_supplies] == I18n.t("coronavirus_form.questions.essential_supplies.options.option_no.label")
+
+    essential_supplies_path
   end
 end
--- a/app/controllers/coronavirus_form/basic_care_needs_controller.rb
+++ b/app/controllers/coronavirus_form/basic_care_needs_controller.rb
@@ -18,18 +18,15 @@ class CoronavirusForm::BasicCareNeedsController < ApplicationController
       respond_to do |format|
         format.html { render controller_path, status: :unprocessable_entity }
       end
-    elsif session[:check_answers_seen]
-      session[:basic_care_needs] = @form_responses[:basic_care_needs]
-      redirect_to check_your_answers_url
     else
       session[:basic_care_needs] = @form_responses[:basic_care_needs]
-      redirect_to dietary_requirements_url
+      redirect_to check_your_answers_url
     end
   end
 
 private
 
   def previous_path
-    essential_supplies_path
+    carry_supplies_path
   end
 end
--- a/app/controllers/coronavirus_form/carry_supplies_controller.rb
+++ b/app/controllers/coronavirus_form/carry_supplies_controller.rb
@@ -18,9 +18,12 @@ class CoronavirusForm::CarrySuppliesController < ApplicationController
       respond_to do |format|
         format.html { render controller_path, status: :unprocessable_entity }
       end
-    else
+    elsif session[:check_answers_seen]
       session[:carry_supplies] = @form_responses[:carry_supplies]
       redirect_to check_your_answers_url
+    else
+      session[:carry_supplies] = @form_responses[:carry_supplies]
+      redirect_to basic_care_needs_url
     end
   end
 
--- a/app/controllers/coronavirus_form/essential_supplies_controller.rb
+++ b/app/controllers/coronavirus_form/essential_supplies_controller.rb
@@ -18,16 +18,27 @@ class CoronavirusForm::EssentialSuppliesController < ApplicationController
       respond_to do |format|
         format.html { render controller_path, status: :unprocessable_entity }
       end
+    elsif @form_responses[:essential_supplies] == I18n.t("coronavirus_form.questions.essential_supplies.options.option_no.label")
+      update_session_store
+      redirect_to dietary_requirements_url
     elsif session[:check_answers_seen]
-      session[:essential_supplies] = @form_responses[:essential_supplies]
+      update_session_store
       redirect_to check_your_answers_url
     else
-      session[:essential_supplies] = @form_responses[:essential_supplies]
-      redirect_to dietary_requirements_url
+      update_session_store
+      redirect_to basic_care_needs_url
     end
   end
 
 private
+
+  def update_session_store
+    session[:essential_supplies] = @form_responses[:essential_supplies]
+    if @form_responses[:essential_supplies] == I18n.t("coronavirus_form.questions.essential_supplies.options.option_yes.label")
+      session[:dietary_requirements] = nil
+      session[:carry_supplies] = nil
+    end
+  end
 
   def previous_path
     nhs_number_path
--- a/app/controllers/coronavirus_form/essential_supplies_controller.rb
+++ b/app/controllers/coronavirus_form/essential_supplies_controller.rb
@@ -18,15 +18,9 @@ class CoronavirusForm::EssentialSuppliesController < ApplicationController
       respond_to do |format|
         format.html { render controller_path, status: :unprocessable_entity }
       end
-    elsif @form_responses[:essential_supplies] == I18n.t("coronavirus_form.questions.essential_supplies.options.option_no.label")
-      update_session_store
-      redirect_to dietary_requirements_url
-    elsif session[:check_answers_seen]
-      update_session_store
-      redirect_to check_your_answers_url
     else
       update_session_store
-      redirect_to basic_care_needs_url
+      redirect_to next_page_url
     end
   end
 
@@ -38,6 +32,18 @@ private
       session[:dietary_requirements] = nil
       session[:carry_supplies] = nil
     end
+  end
+
+  def next_page_url
+    return dietary_requirements_url if answer_next_question?
+    return check_your_answers_url if session[:check_answers_seen]
+
+    basic_care_needs_url
+  end
+
+  def answer_next_question?
+    @form_responses[:essential_supplies] == I18n.t("coronavirus_form.questions.essential_supplies.options.option_no.label") &&
+      (session[:dietary_requirements].blank? || session[:carry_supplies].blank?)
   end
 
   def previous_path
--- a/app/controllers/coronavirus_form/essential_supplies_controller.rb
+++ b/app/controllers/coronavirus_form/essential_supplies_controller.rb
@@ -23,7 +23,7 @@ class CoronavirusForm::EssentialSuppliesController < ApplicationController
       redirect_to check_your_answers_url
     else
       session[:essential_supplies] = @form_responses[:essential_supplies]
-      redirect_to basic_care_needs_url
+      redirect_to dietary_requirements_url
     end
   end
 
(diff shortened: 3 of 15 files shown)


app/controllers/coronavirus_form/carry_supplies_controller.rb
```diff
@@ -18,9 +18,12 @@ def submit
       respond_to do |format|
         format.html { render controller_path, status: :unprocessable_entity }
       end
+    elsif session[:check_answers_seen]
+      session[:essential_supplies] = @form_responses[:essential_supplies]
```
leenagupte: Should this be setting the value for `carry_supplies`?

app/controllers/coronavirus_form/basic_care_needs_controller.rb
```diff
@@ -27,6 +27,6 @@ def submit
 private
 
   def previous_path
-    carry_supplies_path
+    @form_responses[:essential_supplies] == I18n.t("coronavirus_form.questions.essential_supplies.options.option_no.label") ? carry_supplies_path : essential_supplies_path
```
leenagupte: This line is really long. Perhaps it could be split up? 
e.g.

```
return carry_supplies_path if @form_responses[:essential_supplies] == I18n.t("coronavirus_form.questions.essential_supplies.options.option_no.label")

essential_supplies_path
```

app/controllers/coronavirus_form/carry_supplies_controller.rb
```diff
@@ -19,7 +19,7 @@ def submit
         format.html { render controller_path, status: :unprocessable_entity }
       end
     elsif session[:check_answers_seen]
-      session[:essential_supplies] = @form_responses[:essential_supplies]
+      session[:carry_supplies] = @form_responses[:carry_supplies]
```
leenagupte, line 22: Ah! You spotted it. This should probably be fixed up to the previous commit.

spec/controllers/coronavirus_form/essential_supplies_controller_spec.rb
```diff
@@ -35,11 +35,22 @@
...
+
+    it "redirects to next step for a no response" do
+      post :submit, params: { essential_supplies: I18n.t("coronavirus_form.questions.essential_supplies.options.option_no.label") }
       expect(response).to redirect_to(dietary_requirements_path)
     end
 
+    it "clears session values for later food questions for a yes response" do
+      post :submit, params: { essential_supplies: I18n.t("coronavirus_form.questions.essential_supplies.options.option_yes.label") }
+      expect(session[:dietary_requirements]).to be_nil
+      expect(session[:carry_supplies]).to be_nil
```
leenagupte, line 58: Perhaps the test could set some values for `dietary_requirements` and `carry_supplies` before calling POST, so we can be sure that they're actually being cleared.

app/controllers/coronavirus_form/essential_supplies_controller.rb
```diff
@@ -18,12 +18,12 @@ def submit
       respond_to do |format|
         format.html { render controller_path, status: :unprocessable_entity }
       end
+    elsif session[:check_answers_seen] && session[:dietary_requirements].present? && session[:carry_supplies].present?
```
leenagupte: So the user can only go back to `check_answers` directly if they've already answered the `dietary_requirements ` and `carry_supplies` questions? What if they decide not to change their answer to `essential_supplies` and they submit a `yes` again?

spec/controllers/coronavirus_form/essential_supplies_controller_spec.rb
```diff
@@ -57,8 +57,17 @@
...
 
-    it "redirects to check your answers if check your answers previously seen" do
+    it "redirects to to next question if check your answers previously seen and later food questions not answered" do
       session[:check_answers_seen] = true
+      post :submit, params: { essential_supplies: I18n.t("coronavirus_form.questions.essential_supplies.options.option_no.label") }
+
+      expect(response).to redirect_to(dietary_requirements_path)
+    end
+
+    it "redirects to check your answers if check your answers previously seen and later food questions answered" do
```
leenagupte: Can you add a test for, `check_answers_seen` and later food questions not answered, and user answers "yes" to `essential_supplies`, as I think the behaviour needs to be different depending on the answer to that question.

leenagupte (commented):
This is looking really good. I just have a few concerns about the next question logic on the `essential_supplies` page.

I've tried this out locally, and if you answer "yes" to essential supplies the first time, you are routed to the basic needs page and then check your answers. Which is fine.
But if you click change your answers to essential supplies, and then submit "yes" again, you are still routed to the basic care needs page instead of the check your answers page.

app/controllers/coronavirus_form/carry_supplies_controller.rb
brucebolt: Looks like I had fixed this in the wrong commit 🙊

app/controllers/coronavirus_form/essential_supplies_controller.rb
brucebolt: From the check answers page, they can go directly to the later questions if they want to change those questions, since as far as the user is aware these are individual questions rather than linked in some way.  I've only included this logic so they actually get asked the later questions if they are setting the answer to "Yes" for the first time.

spec/controllers/coronavirus_form/essential_supplies_controller_spec.rb
brucebolt: Good spot!  I missed this fundamental part of the test.

app/controllers/coronavirus_form/essential_supplies_controller.rb
leenagupte: I think we've got some crossed wires. If the user previously answered "yes" to `essential_supplies`, completes the form (sees check answers), clicks to change their answer to `essential_supplies`, but then answers "yes" again, they should be routed back to check your answers, not `basic care needs`. Give it a go locally.

brucebolt: Got it!  Was getting my yes and no responses mixed up, until I read the question properly 🤦.

spec/controllers/coronavirus_form/essential_supplies_controller_spec.rb
brucebolt: Added this.  Also picks up the missing logic you have raised in the controller.

brucebolt:
Thanks for the review @leenagupte.  I've addressed all your comments.

leenagupte:
@brucebolt You've got a merge conflict. Sorry about that! Also, I think the schema needs to be updated too.

brucebolt:
I fixed the merge conflict just as you commented 😁.

Will look at the schema after I've had lunch.

spec/controllers/coronavirus_form/essential_supplies_controller_spec.rb
```diff
@@ -46,7 +50,10 @@
     end
 
     it "clears session values for later food questions for a yes response" do
+      session[:dietary_requirements] = "Yes"
+      session[:carry_supplies] = "Yes"
```
leenagupte, line 54: Thanks for adding this. Minor nitpick is that this change should probably be fixed up.

app/controllers/coronavirus_form/essential_supplies_controller.rb
```diff
@@ -18,10 +18,13 @@ def submit
...
+      update_session_store
+      redirect_to check_your_answers_url
     elsif @form_responses[:essential_supplies] == I18n.t("coronavirus_form.questions.essential_supplies.options.option_no.label")
       update_session_store
       redirect_to dietary_requirements_url
-    elsif session[:check_answers_seen]
+    elsif session[:check_answers_seen] && @form_responses[:essential_supplies] == I18n.t("coronavirus_form.questions.essential_supplies.options.option_yes.label")
```
leenagupte: Thanks for fixing the logic. It has made the if/else a bit unwieldy though. Perhaps, rather than having this if/else, the logic for figuring out the next page could be moved to a method?

e.g.

```
  def next_page_url
    return dietary_requirements_url if answer_next_question?
    return check_your_answers_url if session[:check_answers_seen]

    basic_care_needs_url
  end

  def answer_next_question?
    @form_responses[:essential_supplies] == I18n.t("coronavirus_form.questions.essential_supplies.options.option_no.label") &&
      (session[:dietary_requirements].blank? || session[:carry_supplies].blank?)
  end
```

And then the routing logic could be simplified to:

```
    if invalid_fields.any?
      flash.now[:validation] = invalid_fields
      log_validation_error(invalid_fields)

      respond_to do |format|
      end
    else
      update_session_store
      redirect_to next_page_url
    end
```

brucebolt: Nice idea, have made this change.

spec/controllers/coronavirus_form/essential_supplies_controller_spec.rb
brucebolt: Ah, yes, didn't see that I'd added this test in an earlier commit.  Have fixed.

spec/helpers/schema_helper_spec.rb
```diff
@@ -274,6 +274,11 @@
         data = valid_data.merge(dietary_requirements: "Foo")
         expect(validate_against_form_response_schema(data).first).to include("dietary_requirements")
       end
+
+      it "does not return a list of errors when carry_supplies has a nil value" do
```
leenagupte: Typo in test name. Should be `dietary_requirements`.

leenagupte: Also, shouldn't the test on line 268: "returns a list of errors when dietary_requirements is missing" fail now?

That test should probably be replaced with something like:

```
      it "allows nhs_number to be blank" do
        data = valid_data.except(:nhs_number)
        expect(validate_against_form_response_schema(data)).to be_empty	        
      end
```

spec/helpers/schema_helper_spec.rb
```diff
@@ -286,6 +291,11 @@
         data = valid_data.merge(carry_supplies: "Foo")
         expect(validate_against_form_response_schema(data).first).to include("carry_supplies")
       end
+
+      it "does not return a list of errors when carry_supplies has a nil value" do
+        data = valid_data.merge(carry_supplies: nil)
+        expect(validate_against_form_response_schema(data)).to be_empty
+      end
```
leenagupte, line 298: Again here, shouldn't the test on line 290, "returns a list of errors when carry_supplies has an unexpected value" fail as well?

leenagupte (commented):
Almost there. I just have a couple of comments about the schema tests.

spec/helpers/schema_helper_spec.rb
brucebolt: We're not allowing the value to be missing, only to allow it to have a nil value.  Even if the user doesn't answer the later food questions, the value will always be nil: https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/pull/358/files#diff-8c917c651898f2137b6f3b59a44e43f2R32-R33.

leenagupte (approved):
💯 Looks great!

PR #368, opened 2020-05-27
Add check contact details page

leenagupte:
Trello: https://trello.com/c/T7cQOXmQ
Follows on from: PR #345 

Review app 👉  https://coronavirus-check-email-aronzt.herokuapp.com/start

# What's changed?

Adds an additional validation page to ask users to double-check their email address, and adds a suggestion if it looks like the address is incorrect.

# Why?

We are seeing a high incidence of email failures. The hypothesis is that additional validation and suggesting a valid address would improve email delivery rates.

# Expected changes

### With suggestion
Screenshot 2020-05-27 at 14 39 42

### Without suggestion
Screenshot 2020-05-27 at 10 43 22

### Errors
Screenshot 2020-05-27 at 10 43 34

## Check answers page (unchanged)
Screenshot 2020-05-27 at 10 43 00


app/helpers/field_validation_helper.rb
```diff
@@ -93,6 +95,11 @@ def validate_email_address(field, email_address)
     end
   end
 
+  def email_address_suggestion(email_address)
+    suggestion = Mailcheck.new.suggest(email_address)
```
bilbof, line 101: We should guard against `email_address` being nil by returning early, since Mailcheck will throw a `NoMethodError` if `email_address` is nil.

```
NoMethodError
undefined method `downcase' for nil:NilClass
/app/vendor/bundle/ruby/2.6.0/gems/mailcheck-1.0.0/lib/mailcheck.rb:17:in `suggest'
/app/app/helpers/field_validation_helper.rb:99:in `email_address_suggestion'
/app/app/controllers/coronavirus_form/check_contact_details_controller.rb:6
```

This can happen if you visit `/check-contact-details` without an email address in your contact details hash.

bilbof (approved):
Great work 👍 just a minor suggestion

app/helpers/field_validation_helper.rb
leenagupte: Thanks @bilbof! I've updated and fixed up.

bilbof (approved):

PR #375, opened 2020-06-01
Update definition of shielding

StephenGill:
Update definition of shielding to bring it into line with latest government guidance.

https://trello.com/c/OvnXWwW7/533-updated-start-page-to-be-consistent-with-new-shielding-guidance


issyl0 (approved):
Thanks!

PR #376, opened 2020-06-01
Guard against session contact_details being unset

bilbof:
This is unexpected behaviour, but should prevent a user from seeing a 500 error if they don't have contact_details set.

Fixes https://sentry.io/organizations/govuk/issues/1703071294


alex-ju (approved):
🐛🔨

PR #380, opened 2020-06-02
Move the nhs number question

JonathanHallam:
This moves the NHS number question to immediately after the questions establishing whether a person is eligible. This is because the NHS number is something that people may not have immediately at hand so needs to be at the start of the flow.

### DIFF
--- a/spec/features/filling_in_the_form/fill_in_the_form_spec.rb
+++ b/spec/features/filling_in_the_form/fill_in_the_form_spec.rb
@@ -10,12 +10,12 @@ RSpec.feature "fill in the vulnerable people form" do
       given_an_extremely_vulnerable_person_during_the_covid_19_pandemic
       that_lives_in_england
       and_has_recently_received_an_nhs_letter
+      and_has_given_their_nhs_number
       and_has_given_their_name
       and_has_given_their_date_of_birth
       and_has_given_their_address
       and_has_given_their_contact_details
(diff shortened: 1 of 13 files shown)


spec/features/filling_in_the_form/fill_in_the_form_spec.rb
```diff
@@ -10,12 +10,12 @@
       given_an_extremely_vulnerable_person_during_the_covid_19_pandemic
       that_lives_in_england
       and_has_recently_received_an_nhs_letter
+      and_has_given_their_nhs_number
       and_has_given_their_name
       and_has_given_their_date_of_birth
       and_has_given_their_address
       and_has_given_their_contact_details
       and_has_checked_their_contact_details
-      and_has_given_their_nhs_number
```
bilbof, line 18: This is a really good diff for what changed :)

bilbof (approved):
Looks good to me 👍 

As usual it's tricky to think about how to handle users currently active on the form when changing questions... E.g. in this case if someone is on the form, but hasn't yet given an NHS number, they might not see the NHS number page. There's not much one can do to reduce user impact for now, aside from merging this outside of peak hours.

There's probably a better way we can design the routing so we can do changes like this a bit more safely.. there's probably another card in this!

gclssvglx (approved):
LGTM 👍

PR #381, opened 2020-06-02
Add UPRN to schema

bilbof:
Also removes postcode from top-level of session.

Fixes https://sentry.io/organizations/govuk/issues/1667664108

https://trello.com/c/Xug1GjUB/536-add-new-uprn-and-postcode-fields-to-schema

### DIFF
--- a/app/controllers/coronavirus_form/address_lookup_controller.rb
+++ b/app/controllers/coronavirus_form/address_lookup_controller.rb
@@ -18,6 +18,7 @@ class CoronavirusForm::AddressLookupController < ApplicationController
 
     address = JSON.parse(params[:address]).to_h
     session[:support_address] = helpers.convert_address(address)
+    session.delete(:postcode)
     redirect_to support_address_path
   rescue JSON::ParserError, AddressNotProvidedError
     render controller_path, status: :unprocessable_entity
--- a/config/schemas/form_response.json
+++ b/config/schemas/form_response.json
@@ -97,6 +97,11 @@
         "postcode": {
           "type": "string",
           "pattern": "^[A-Za-z0-9 ]{4,}$"
+        },
+        "uprn": {
+          "type": ["string", "null"],
+          "maxLength": 12,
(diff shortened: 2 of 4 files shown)


app/controllers/coronavirus_form/address_lookup_controller.rb
```diff
@@ -18,6 +18,7 @@ def submit
 
     address = JSON.parse(params[:address]).to_h
     session[:support_address] = helpers.convert_address(address)
+    session.delete(:postcode)
```
leenagupte, line 21: Why is this being deleted?

config/schemas/form_response.json
```diff
@@ -97,6 +97,9 @@
         "postcode": {
           "type": "string",
           "pattern": "^[A-Za-z0-9 ]{4,}$"
+        },
+        "uprn": {
+          "type": ["string", "null"]
```
leenagupte: It looks like UPRNs are always 12 digits, so do we want to add a pattern matcher here?

app/controllers/coronavirus_form/address_lookup_controller.rb
bilbof: This is added [here](https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/blob/290adc4f9e2def85ab7ca32c1c7ba3c105c829d1/app/controllers/coronavirus_form/postcode_lookup_controller.rb#L26) but only for the address lookup. This is a duplicate of the `support_address.postcode` field.

config/schemas/form_response.json
bilbof: Yeah good point. Jon and Graham tell me they can be under 12 digits, so I've added a `"maxLength"` of 12 there.

leenagupte: And limiting it to numbers too?

bilbof: Thanks :-)

leenagupte (approved):
🎉

PR #384, opened 2020-06-03
Codecov GitHub action

Rosa-Fox:
What
----

- Codecov is a code coverage reporting tool. It checks that your tests still test the same (or more) of your code when you add new code. This is an example report from the docs: https://codecov.io/gh/codecov/example-ruby.

- Codecov is free for public repos.

- This will enable us to identify any gaps in our test coverage.

- There is a Github Marketplace app (https://github.com/marketplace/codecov) for Codecov which has already been configured for all public AlphaGov repositories.

- There is also a Github Marketplace Actions app (https://github.com/marketplace/actions/codecov) for Codecov.

- Reports need to be in XML format to be uploaded to Codecov, so there is a commit to generate the coverage report as XML.

- Uploading the report to Codecov as a Github Action is configured within this PR.

How to review
-------------

- Have a look at https://github.com/marketplace/actions/codecov and see if you think this looks ok in comparison. 

- Its hard to test if it has worked until it runs... once it has then:

1. Go to: https://codecov.io 
2. Sign in with Github
3. Choose the govuk-coronavirus-vulnerable-people-form repo from the menu
4. You should see some metrics
5. A bot also posts comments in Github

Links
-----

[Trello](https://trello.com/c/iKWfHb9n/437-add-codecov-to-our-ci-pipelines)


.github/workflows/ci.yml
```diff
@@ -16,3 +16,7 @@ jobs:
         restore-keys: bundle
     - run: bundle install --jobs 4 --retry 3 --deployment
     - run: bundle exec rake
+    - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v1.0.7
+        with:
+          file: ./coverage/coverage.xml
```
issyl0:
```suggestion
      uses: codecov/codecov-action@v1.0.7
      with:
        file: ./coverage/coverage.xml
```

This was erroring: https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/actions/runs/123839661

issyl0 (changes requested):

.github/workflows/ci.yml
Rosa-Fox: 😱 have updated!

issyl0 (approved):

PR #388, opened 2020-06-04
Deactive HTML5 validation on email

bilbof:
This makes the email input consistent with the other place we ask
users for input. We have server side validation for this, and I'm
told HTML5 validation isn't reliable.

https://trello.com/c/6wyg1U0O/549-remove-html5-validation-from-new-check-email-page


alex-ju (approved):
Ha, great catch. I assume the tracking wasn't working either. 🎖️

bilbof:
Thx, yeah it was being treated as a query parameter.

PR #389, opened 2020-06-04
check_contact_details: Handle `nil`  when finding email address

issyl0:
What
----

Handle `nil` contact details when trying to find email address

- We were seeing [Sentry errors](https://sentry.io/organizations/govuk/issues/1709557316/?project=5170680&referrer=slack):

```
undefined method `dig' for nil:NilClass
@email_suggestion = email_address_suggestion(@form_responses[:contact_details].dig(:email))
```

- We don't enforce that users must give us contact details (including email address) - because they might not have any of these details.

How to review
-------------

Observe that your local form doesn't get errors?

Links
-----

https://sentry.io/organizations/govuk/issues/1709557316/?project=5170680&referrer=slack


alex-ju (approved):
💯

PR #390, opened 2020-06-05
Preserve user-provided name fields in name view

bilbof:
This fixes a bug, where a user wouldn't be shown the names they had entered thanks to us using Strings instead of Symbols.

https://trello.com/c/AQs4RW9M/547-if-you-error-on-name-page-it-loses-any-data-you-entered-on-that-page


leenagupte (approved):
🥇 Love the test!

PR #392, opened 2020-06-05
Guard against support_address being undefined

bilbof:
Resolves https://sentry.io/organizations/govuk/issues/1711254097


leenagupte (approved):
👍

PR #394, opened 2020-06-08
Remove additional 'postcode' field from FormResponse

bilbof:
This field is put into the session in postcode lookup controller but the data pipeline uses the support_address.postcode field.

This fixes a validation error: https://sentry.io/organizations/govuk/issues/1667664108

### DIFF
--- a/app/controllers/coronavirus_form/check_answers_controller.rb
+++ b/app/controllers/coronavirus_form/check_answers_controller.rb
@@ -39,6 +39,8 @@ class CoronavirusForm::CheckAnswersController < ApplicationController
     redirect_to confirmation_url(reference_number: submission_reference, contact_gp: @contact_gp)
   end
 
+  EXCLUDED_FIELDS = %i[session_id _csrf_token current_path previous_path check_answers_seen postcode].freeze
+
 private
 
   def smoke_tester?
@@ -94,7 +96,7 @@ private
   end
 
   def sanitised_session
-    session_with_indifferent_access.except(:session_id, :_csrf_token, :current_path, :previous_path, :check_answers_seen)
+    session_with_indifferent_access.except(*EXCLUDED_FIELDS)
   end
 
   def contact_gp?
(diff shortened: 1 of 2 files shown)


app/controllers/coronavirus_form/check_answers_controller.rb
```diff
@@ -39,6 +39,8 @@ def submit
     redirect_to confirmation_url(reference_number: submission_reference, contact_gp: @contact_gp)
   end
 
+  EXCLUDED_FIELDS = %i[session_id _csrf_token current_path previous_path check_answers_seen postcode].freeze
```
leenagupte, line 42: Should `postcode` be removed from the session in postcode lookup controller instead? Can `support_address.postcode` be used for all of the lookups instead?

spec/controllers/coronavirus_form/check_answers_controller_spec.rb
```diff
@@ -162,6 +166,25 @@
...
+    context "when given excluded fields" do
+      before do
+        @submitted_data = valid_data.merge(reference_id: "abc")
+        session.merge!(@submitted_data)
+        described_class::EXCLUDED_FIELDS.each do |field|
+          session[field.to_sym] = "bad_value"
+        end
+      end
+
+      it "excluded fields are removed before creating the FormResponse" do
```
leenagupte, line 178:
```suggestion
      it "removes excluded fields before creating the FormResponse" do
```

leenagupte (commented):
The code looks good. I just have one question about why the postcode isn't being removed from the session earlier. I'm also still not clear why `support_address.postcode` isn't being used in all of the lookups as we don't need to worry about that field being in the session.

spec/controllers/coronavirus_form/check_answers_controller_spec.rb
```diff
@@ -41,6 +41,10 @@
         ENV["HEROKU_APP_NAME"] = "coronavirus-form-preview"
       end
 
+      after do
+        ENV["HEROKU_APP_NAME"] = nil
+      end
+
```
GDSNewt, line 47: Hi Bill, i'm not really familiar with the use of  `after do` does it just execute that bit of code when the `context "when running as a Heroku preview app" do ` ends?

app/controllers/coronavirus_form/check_answers_controller.rb
bilbof: You are right, I will close this and change the controller in another PR. We should probably also go from excluding fields from the session to including them, perhaps using keys from the schema? :)

bilbof:
Superseded by https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/pull/397.

PR #395, opened 2020-06-08
Add brakeman

bilbof:
[Brakeman](https://github.com/presidentbeef/brakeman) enables us to do static analysis of the codebase for security issues. It'll run as part of the default rake task or simply `bundle exec brakeman`.


gclssvglx (approved):
Good move 👍

PR #396, opened 2020-06-08
Fix bug with passing on UPRN

gclssvglx:
What
----

When the user moves part of the chosen address around (from the County
to the 2nd address line for example) the UPRN can get removed.  We want
to maintain the UPRN as much as possible.

This change ensures the UPRN is maintained against changes that do not
directly effect the postcode or other fundamental address elements. This
should ensure that we pass on more UPRN's.

Links
-----

[Trello card](https://trello.com/c/g2Jhll4F/566-fix-bug-with-passing-on-uprn)

### DIFF
--- a/app/helpers/address_helper.rb
+++ b/app/helpers/address_helper.rb
@@ -58,7 +58,6 @@ module AddressHelper
       town_city: address_line_builder(raw_address, TOWN_CITY).titleize,
       county: address_line_builder(raw_address, COUNTY).titleize,
       postcode: address_line_builder(raw_address, POSTCODE).delete(" "),
-      sao_present: sao.present?,
     }
 
     if sao.present?
@@ -86,19 +85,40 @@ module AddressHelper
     address.slice(*WANTED_VALUES)
   end
 
-  def update_support_address(selected, given)
-    given[:uprn] = selected[:uprn] if compare(selected, given)
-    given.delete(:sao_present)
-    { support_address: given }
+  def update_support_address(ordnance_address, edited_address)
+    edited_address[:uprn] = ordnance_address[:uprn] if compare(ordnance_address, edited_address)
+    { support_address: edited_address }
   end
 
-  def compare(selected, given)
-    return false if selected.empty?
+  def sanitize_postcode(postcode)
+    postcode.gsub(/[^0-9a-zA-Z]+/, "")
+  end
 
-    line_one_and_two = address_comparison_line_builder(selected, LINE_1_2_AND_POSTCODE) == address_comparison_line_builder(given, LINE_1_2_AND_POSTCODE)
+  def sanitize_address(address)
+    return [] if address.blank?
 
-    line_one = address_comparison_line_builder(selected, LINE_1_AND_POSTCODE) == address_comparison_line_builder(given, LINE_1_AND_POSTCODE)
+    if address.values.any?
+      copy_of_address = address
+      copy_of_address[:postcode] = sanitize_postcode(address[:postcode])
 
-    selected[:sao_present] ? line_one_and_two : line_one_and_two || line_one
+      copy_of_address
+        .with_indifferent_access
+        .except(:uprn, :county, :town_city)
+        .values
+        .map(&:upcase).join(" ")
+        .gsub(/[^0-9A-Z]+/, " ")
+        .split(" ")
+        .sort
+        .compact
+    end
+  end
+
+  def compare(ordnance_address, edited_address)
+    return false if ordnance_address.blank?
+
+    sanitized_ordnance_address = sanitize_address(ordnance_address)
+    sanitized_edited_address = sanitize_address(edited_address)
+
+    sanitized_edited_address & sanitized_ordnance_address == sanitized_ordnance_address
   end
 end
--- a/spec/helpers/address_helper_spec.rb
+++ b/spec/helpers/address_helper_spec.rb
@@ -112,108 +112,153 @@ RSpec.describe AddressHelper, type: :helper do
   end
 
   describe "#update_support_address" do
-    it "updates the UPRN when the selected address compares to the given address" do
-      selected = {
+    it "updates the UPRN when the ordnance_address address compares to the edited_address address" do
+      ordnance_address = {
         uprn: "123456789",
         building_and_street_line_1: "address line 1",
         postcode: "1234 567",
       }
 
-      given = {
+      edited_address = {
         building_and_street_line_1: "address line 1",
         postcode: "1234 567",
       }
 
-      expect(helper.update_support_address(selected, given).dig(:support_address, :uprn)).to eq("123456789")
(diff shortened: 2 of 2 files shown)


app/helpers/address_helper.rb
```diff
@@ -88,17 +87,38 @@ def make_address_hash(address)
...
   def update_support_address(selected, given)
     given[:uprn] = selected[:uprn] if compare(selected, given)
-    given.delete(:sao_present)
     { support_address: given }
   end
 
-  def compare(selected, given)
-    return false if selected.empty?
+  def sanitize_postcode(postcode)
+    postcode.gsub(/[^0-9a-zA-Z]+/, "")
```
leenagupte, line 94: Is this doing the same as `postcode = ""` ?

app/helpers/address_helper.rb
```diff
@@ -88,17 +87,38 @@ def make_address_hash(address)
...
+        .map(&:upcase).join(" ")
+        .gsub(/[^0-9A-Z]+/, " ")
+        .split(" ")
+        .sort
+        .compact
+    end
+  end
 
-    line_one_and_two = address_comparison_line_builder(selected, LINE_1_2_AND_POSTCODE) == address_comparison_line_builder(given, LINE_1_2_AND_POSTCODE)
+  def compare(selected, given)
```
leenagupte: I'm not sure what `selected` and `given` mean here. Can they be given more meaningful names?

app/helpers/address_helper.rb
```diff
@@ -88,17 +87,38 @@ def make_address_hash(address)
...
+    return false if selected.blank?
```
leenagupte: What happens if "given" is false?

app/helpers/address_helper.rb
```diff
@@ -88,17 +87,38 @@ def make_address_hash(address)
...
 
-    line_one = address_comparison_line_builder(selected, LINE_1_AND_POSTCODE) == address_comparison_line_builder(given, LINE_1_AND_POSTCODE)
+    a = sanitize_address(selected)
+    b = sanitize_address(given)
 
-    selected[:sao_present] ? line_one_and_two : line_one_and_two || line_one
+    b & a == a
```
leenagupte: Why is the union being checked? 

I think it might clearer if I knew what "a" and "b" are supposed to represent, and why common elements of a and b must equal a. Why not b?

spec/helpers/address_helper_spec.rb
```diff
@@ -128,92 +128,122 @@
...
     end
 
-    context "SAO is false" do
-      it "returns true if first line and postcode are equivalent" do
-        selected = {
-          sao_present: false,
-          building_and_street_line_1: "10 WHITECHAPEL HIGH STREET",
-          postcode: "E18QS",
-        }
+    it "returns true if anything is moved but not removed" do
```
leenagupte, line 199: Great test! It really clears up what the method is doing. I still think `selected` and `given` should be renamed. Perhaps "edited_address" and "ordnance_address"?

spec/helpers/address_helper_spec.rb
```diff
@@ -128,92 +128,122 @@
...
+      expect(helper.compare(selected, given)).to be true
+    end
 
-      it "returns false if the postcodes do not match" do
-        selected = {
-          sao_present: false,
-          building_and_street_line_1: "10 WHITECHAPEL HIGH STREET",
-          postcode: "E18QS",
-        }
+    it "returns false if anything has been removed building_and_street_line_1" do
```
leenagupte:
```suggestion
    it "returns false if anything has been removed from building_and_street_line_1" do
```

spec/helpers/address_helper_spec.rb
```diff
@@ -128,92 +128,122 @@
...
+      expect(helper.compare(selected, given)).to be false
+    end
 
-      it "returns false if the first lines do not match" do
-        selected = {
-          sao_present: false,
-          building_and_street_line_1: "10 WHITECHAPEL HIGH STREET",
-          postcode: "E18QS",
-        }
+    it "returns false if anything has been changed" do
```
leenagupte:
```suggestion
    it "returns false if the postcode has been changed" do
```

leenagupte (commented):
The tests look great 🎉 

I've just got a couple of questions about the logic, most of which I think could be cleared up by renaming some of the variables.

JonathanHallam (commented):
I'm happy with the changes. I think I might be a bit close to give ultimate approval but confident that it's the best way to go in terms of logic and that it fixes the bug.

app/helpers/address_helper.rb
gclssvglx: Hi @leenagupte - thanks for the review - fantastic suggestions 👍  Essentially, this is removing anything that is not (^) an alphanumeric character.  So, something like "_ AA1A & 1AA -" would become "AA1A1AA".  This is done to maintain the postcode's integrity.

app/helpers/address_helper.rb
gclssvglx: Yes, rubbish names - sorry.  Fixed throughout.  BTW - love your suggested names - they work well 👍 .

app/helpers/address_helper.rb
gclssvglx: The expectation is that we would get false.  The reason for not testing it explicitly is because the compare will handle that, and we're really only tested that the user has selected an address at this point.

app/helpers/address_helper.rb
gclssvglx: Fixed.

gclssvglx: So, the reason for the test here is we want to ensure that everything (that we're interested in) is maintained in the edited address.

spec/helpers/address_helper_spec.rb
gclssvglx: Thanks 👍 - names fixed.

app/helpers/address_helper.rb
leenagupte: I realised once you'd renamed everything 😄

spec/helpers/address_helper_spec.rb
gclssvglx: Whoops - fixed - thanks

spec/helpers/address_helper_spec.rb
gclssvglx: Good suggestion - thanks - fixed

app/helpers/address_helper.rb
gclssvglx: Good point - I've added explicit tests for either the edited or ordnance addresses being blank.

gclssvglx:
Thanks @leenagupte and @JonathanHallam for the reviews, comments and suggestions 👍

leenagupte (approved):
LGTM! 💯  Thanks for updating the variable names. The new names along with the tests make this really easy to follow.

PR #397, opened 2020-06-08
Remove duplicate postcode field from session

bilbof:
We store postcode in `session[:postcode]` for the postcode lookup,
but we can use `session[:support_address][:postcode]` instead, so
that we don't later have to remove this field from the session.


leenagupte (approved):
This is nice! 🎉

PR #399, opened 2020-06-09
Add docs about how to deploy breaking changes

leenagupte:
Trello: https://trello.com/c/2nrSdb3X

# What's changed and why?
Adds docs about how to deploy breaking changes so that we don't have ask
people to deploy things out of hours.

## Other approaches considered
Before settling on just adding guidance, pinning the user to a specific
version of the application or the JSON schema was considered.

However because we use continuous deployment to deploy the app, there
would be an added overhead of updating the application version number
for all changes except content changes. Also it doesn't look like it's
possible to add a version property to a JSON schema yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,6 +91,10 @@ which is configured in [concourse/pipeline.yml](concourse/pipeline.yml).
 The concourse pipeline has credentials for the `govuk-forms-deployer` user in
 GOV.UK PaaS. This user has the SpaceDeveloper role, so it can `cf push` the application.
 
+## How to deploy breaking changes
+
+Details can be found [here](docs/how-to-deploy-breaking-changes).
+
 ## Licence
 
 [MIT License](LICENCE)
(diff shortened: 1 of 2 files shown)


OllieJC:
Is this an appropriate place to add info about impact downstream too? e.g.: field removals from schema are _fine_ but renames or additional fields won't get pulled through automatically and delivered to consumers (LAs/wholesalers)

leenagupte:
> Is this an appropriate place to add info about impact downstream too? e.g.: field removals from schema are _fine_ but renames or additional fields won't get pulled through automatically and delivered to consumers (LAs/wholesalers)

@OllieJC I was thinking about "breaking changes" in terms of the users of the form, but yes, I think it's valuable to add the impact on downstream data as a consideration. I can add a note to remind people to notify the data team if they are making a breaking data change.

bilbof (approved):
Nice one, thank you for this. I enjoyed the reference to A AB B deployments.

leenagupte:
@OllieJC @bilbof Thanks for your reviews. I've added a paragraph about remembering to update the downstream data team about any breaking changes. I haven't added contact details because I'm not sure how appropriate it is to add a slack channel to a README.

PR #400, opened 2020-06-10
Stop the potential for non-string data

gclssvglx:

This change removes all non-String data from the entered form data, prior to being saved.  Note that `instance_of?(String)` is used - instead of `is_a?` or `kind_of?` as it checks only the direct Object class and not any super-type(s) it might have.  So, this would also remove any data who's class is a sub-type of String.

Links
-----
[Incident Google Doc](https://docs.google.com/document/d/16qU8ZgSI7Y_IWN7F1nCkLWtVFvX_jw8VWwAiT9rs7EI/edit#heading=h.p99426yo0rbv)
[Incident investigation Trello card](https://trello.com/c/U8FR01EG/581-investigate-root-cause-of-incident)

### DIFF
--- a/app/helpers/address_helper.rb
+++ b/app/helpers/address_helper.rb
@@ -105,8 +105,9 @@ module AddressHelper
         .with_indifferent_access
         .except(:uprn, :county, :town_city)
         .values
-        .map(&:to_s)
-        .map(&:upcase).join(" ")
+        .select { |value| value.instance_of?(String) }
+        .map(&:upcase)
+        .join(" ")
         .gsub(/[^0-9A-Z]+/, " ")
         .split(" ")
         .sort
--- a/spec/helpers/address_helper_spec.rb
+++ b/spec/helpers/address_helper_spec.rb
@@ -144,24 +144,21 @@ RSpec.describe AddressHelper, type: :helper do
       expect(helper.sanitize_address(address)).to eq array
     end
 
-    it "returns a valid array if part of the address is nil" do
+    it "returns an array of strings if the address contains non-string values" do
       address = {
-        one: "one two three 1234567890",
-        two: nil,
+        string: "one",
+        boolean: true,
+        null: nil,
+        integer: 100,
+        float: 1.34,
+        object: Object.new,
+        struct: Struct.new("Ignored"),
+        array: [],
+        hash: {},
+        symbol: :ignored,
         postcode: "AA1A 1AA",
       }
-      array = %w[1234567890 AA1A1AA ONE THREE TWO]
-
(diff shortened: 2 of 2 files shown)


spec/helpers/address_helper_spec.rb
```diff
@@ -144,24 +144,21 @@
       expect(helper.sanitize_address(address)).to eq array
     end
 
-    it "returns a valid array if part of the address is nil" do
+    it "returns an array of strings if the address contains non-string values" do
       address = {
-        one: "one two three 1234567890",
-        two: nil,
+        string: "one",
```
bilbof, line 149: Would it help the test if the bad values were put into keys that the helper is expecting like `uprn` or `building_and_street_line_1`, rather than `symbol:`?

```suggestion
        building_and_street_line_1: "one",
        building_and_street_line_2: 100,
        town_city: 1.34,
        county: Object.new,
        uprn: Struct.new("Ignored"),
        array: [],
        postcode: "AA1A 1AA",
```

app/helpers/address_helper.rb
```diff
@@ -105,8 +105,9 @@ def sanitize_address(address)
         .with_indifferent_access
         .except(:uprn, :county, :town_city)
         .values
-        .map(&:to_s)
-        .map(&:upcase).join(" ")
+        .select { |value| value.instance_of?(String) }
```
bilbof, line 108: This is a good change that'll make the code more robust 👍 

I wonder why `value` here can be something other than a `String`? 🤔

bilbof (approved):

spec/helpers/address_helper_spec.rb
gclssvglx: Can do, but there are some reasons why I don't feel this is necessary or useful:

* It is the `sanitize_address` method that was throwing the `NoMethodError`.  This method doesn't care about the format of the address, but it does remove UPRN, country and town_city. 

* There are only 5 fields in the support_address - and 3 of those are removed by the `sanitize_method`, so I'd need a number tests to test all types of bad value.

Happy to add if you feel testing the `sanitize_method` needs these specific fields.

bilbof: Thanks for explaining. This makes sense :shipit:

app/helpers/address_helper.rb
gclssvglx: At the moment - all of the form inputs are `type=text`.  So, not only will they be converted by the browser into string params, but we do not want anything other than a string in them.  So, I feel removing anything but a string is reasonable at the moment.

gclssvglx: Of course, this may change in the future!

PR #406, opened 2020-06-11
Address incident recommendations

gclssvglx:

This change guards against `nil` fields which can be created by calling `presence` on `nil`, and removes any changes that may have been made to the selected address (the address that has been chosen from the address lookup results dropdown).  The `UPRN` field is additionally removed if it can no longer be trusted.

Links
-----

[Incident Google Doc](https://docs.google.com/document/d/16qU8ZgSI7Y_IWN7F1nCkLWtVFvX_jw8VWwAiT9rs7EI/edit#heading=h.p99426yo0rbv)
[Incident investigation Trello card](https://trello.com/c/U8FR01EG/581-investigate-root-cause-of-incident)

### DIFF
--- a/app/controllers/coronavirus_form/support_address_controller.rb
+++ b/app/controllers/coronavirus_form/support_address_controller.rb
@@ -56,12 +56,20 @@ private
 
   def given_address
     {
-      building_and_street_line_1: strip_tags(params[:building_and_street_line_1]&.strip).presence,
-      building_and_street_line_2: strip_tags(params[:building_and_street_line_2]&.strip).presence,
-      town_city: strip_tags(params[:town_city]&.strip).presence,
-      county: strip_tags(params[:county]&.strip).presence,
-      postcode: strip_tags(params[:postcode]&.gsub(/[[:space:]]+/, "")).presence,
+      building_and_street_line_1: strip_tags_from_address_field(params[:building_and_street_line_1]),
+      building_and_street_line_2: strip_tags_from_address_field(params[:building_and_street_line_2]),
+      town_city: strip_tags_from_address_field(params[:town_city]),
+      county: strip_tags_from_address_field(params[:county]),
+      postcode: strip_tags_from_postcode_field(params[:postcode]),
     }
+  end
+
+  def strip_tags_from_address_field(param)
+    param.nil? ? "" : strip_tags(param&.strip).presence
+  end
+
+  def strip_tags_from_postcode_field(postcode)
+    postcode.nil? ? "" : strip_tags(postcode&.gsub(/[[:space:]]+/, "")).presence
   end
 
   def validate_fields
--- a/app/helpers/address_helper.rb
+++ b/app/helpers/address_helper.rb
@@ -123,4 +123,12 @@ module AddressHelper
 
     sanitized_edited_address & sanitized_ordnance_address == sanitized_ordnance_address
   end
+
+  def remove_changes_to_ordnance_survey_api_response(ordnance_address)
+    address = JSON.parse(ordnance_address).to_h
+
+    address.select do |key, value|
+      WANTED_VALUES.include?(key) && value.instance_of?(String)
+    end
+  end
 end
--- a/app/controllers/coronavirus_form/address_lookup_controller.rb
+++ b/app/controllers/coronavirus_form/address_lookup_controller.rb
@@ -16,7 +16,7 @@ class CoronavirusForm::AddressLookupController < ApplicationController
   def submit
     raise AddressNotProvidedError if params[:address].blank?
 
-    address = JSON.parse(params[:address]).to_h
+    address = helpers.remove_changes_to_ordnance_survey_api_response(params[:address])
     session[:support_address] = helpers.convert_address(address)
     redirect_to support_address_path
   rescue JSON::ParserError, AddressNotProvidedError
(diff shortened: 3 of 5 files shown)


app/helpers/address_helper.rb
```diff
@@ -123,4 +123,19 @@ def compare(ordnance_address, edited_address)
...
+    address_hash = JSON.parse(selected_address).to_h
+    selected_address_keys = address_hash.keys
+
+    address_to_return = address_hash.select do |key, value|
+      WANTED_VALUES.include?(key) && value.instance_of?(String)
+    end
+
+    unless address_to_return.keys == selected_address_keys
+      address_to_return.delete("UPRN")
+    end
```
bilbof: I don't think this block is necessary, since non-malicious users won't modify the address, and therefore the UPRN doesn't matter for these users.

app/helpers/address_helper.rb
```diff
@@ -123,4 +123,19 @@ def compare(ordnance_address, edited_address)
...
+
+    address_to_return
```
bilbof: I think validating that we're receiving the data that we expect is a good idea 👍 

This method currently wouldn't prevent users from submitting invalid values to be stored in the session. We would need to validate the data submitted against the options we had provided. I.e. cache the valid options received from the API, and then compare the user provided data with the cached options.

app/helpers/address_helper.rb
```diff
@@ -123,4 +123,19 @@ def compare(ordnance_address, edited_address)
 
     sanitized_edited_address & sanitized_ordnance_address == sanitized_ordnance_address
   end
+
+  def remove_selected_address_changes(selected_address)
```
leenagupte:
```suggestion
  def remove_changes_to_ordnance_survey_api_response(ordnance_address)
```

leenagupte (commented):
@gclssvglx I'm a little confused some of the changes in this PR. 

In the first commit, why are `nils` being replaced with empty strings? Can the code using the output from `given_address` skip keys with `nil` values?

The tests really help to make clear what's happening in the second commit ⭐ . Thank you! Can you update the commit message to say why it's import to remove any extra keys that might have been added to the api response. Also, what happens if the ordnance survey adds new keys to the response?

app/helpers/address_helper.rb
gclssvglx: Thanks @bilbof - I have removed this block as requested.  However, personally I disagree - it could be seen that if we have a valid UPRN then the rest of the address can reasonably be ignored and the entry assumed to be from a real user.  Removing the UPRN - when we know that someone has munged the address - is a safe-guard against this assumption.

app/helpers/address_helper.rb
gclssvglx: Thanks @leenagupte - like the name.  Updated.

app/helpers/address_helper.rb
gclssvglx: Thanks @bilbof - perhaps we can have a brief catch up on this?

gclssvglx:
Hi @leenagupte - thanks for the comments and suggestions 👍 

Nils are replaced with empty strings as this is how the `nil` value is getting into the data. Yes, we could skip fields with nil values later, but it's likely that would happen after the address has already entered the session store. That just messes up the data downstream.

> what happens if the ordnance survey adds new keys to the response?

Nothing, I'd hope! We are currently using `v1` of the API, adding fields wouldn't necessarily be seen as a breaking change, but changing or renaming an existing field definitely would be.  Any new fields, by definition, means they are non essential to the `v1` functionality.  If the new fields change the core functionality - I'd expect them to be a `v2` update - in which case we'd have to move to `v2` when possible.

bilbof (approved):

app/helpers/address_helper.rb
bilbof: I think it's not something we'd need to do in this PR, so we can do it later if needed 👍

leenagupte:
> Nils are replaced with empty strings as this is how the `nil` value is getting into the data. Yes, we could skip fields with nil values later, but it's likely that would happen after the address has already entered the session store. That just messes up the data downstream.

@gclssvglx Thanks for the explanation! Could you add it to the commit note as well?

app/controllers/coronavirus_form/support_address_controller.rb
```diff
@@ -56,14 +56,22 @@ def set_default_support_address
...
+      building_and_street_line_1: strip_tags_from_address_field(params[:building_and_street_line_1]),
+      building_and_street_line_2: strip_tags_from_address_field(params[:building_and_street_line_2]),
+      town_city: strip_tags_from_address_field(params[:town_city]),
+      county: strip_tags_from_address_field(params[:county]),
+      postcode: strip_tags_from_postcode_field(params[:postcode]),
     }
   end
 
+  def strip_tags_from_address_field(param)
+    param.nil? ? "" : strip_tags(param&.strip).presence
```
leenagupte, line 68: I think you might still end up with nils. for example:

```
[8] pry(main)> a = "<p>"
=> "<p>"
[9] pry(main)> strip_tags(a)
=> ""
[10] pry(main)> strip_tags(a).presence
=> nil
```

gclssvglx: Damn!  Thanks @leenagupte I'll see if I can fix that.

PR #407, opened 2020-06-11
Bug - postcode not showing

gclssvglx:
Postcode is not visible to the left of the word 'Change' on the
address lookup page. It should be.

This change corrects the session hash lookup to be a string key
lookup and not symbol key.

### DIFF
--- a/app/views/coronavirus_form/address_lookup.html.erb
+++ b/app/views/coronavirus_form/address_lookup.html.erb
@@ -24,7 +24,9 @@
 } %>
 
 <p class="govuk-body">
-  <span class="govuk-!-font-weight-bold govuk-!-padding-right-3"><%= session.dig(:support_address, :postcode) %></span>
+  <span class="govuk-!-font-weight-bold govuk-!-padding-right-3">
+    <%= session.to_h.with_indifferent_access.dig(:support_address, :postcode) %>
+  </span>
   <%= link_to "Change", previous_path, class: "govuk-link" %>
 </p>
 


app/views/coronavirus_form/address_lookup.html.erb
```diff
@@ -24,7 +24,7 @@
 } %>
 
 <p class="govuk-body">
-  <span class="govuk-!-font-weight-bold govuk-!-padding-right-3"><%= session.dig(:support_address, :postcode) %></span>
+  <span class="govuk-!-font-weight-bold govuk-!-padding-right-3"><%= session.dig(:support_address, "postcode") %></span>
```
leenagupte: I think other places we've gotten around this by doing:

```
session.to_h.with_indifferent_access.dig(:support_address, :postcode)
```

leenagupte: It might be easier to access `@addresses` instead which already has `with_indifferent_access` applied to it. i.e. `@addresses.dig(:support_address, :postcode)`

gclssvglx: Thanks for the review and the suggestions @leenagupte 👍  

Good shout - I like this option...  

> I think other places we've gotten around this by doing:
> 
> ```
> session.to_h.with_indifferent_access.dig(:support_address, :postcode)
> ```

The other option you mention wouldn't be so easy as each address in `@addresses` looks like this... 

```
{
  :text => "STREET RECORD, AMBASSADORS COURT, LONDON, CITY OF WESTMINSTER, SW1A 1AA", 
  :value => "{\"UPRN\":\"10033555818\",\"PAO_TEXT\":\"STREET RECORD\",\"STREET_DESCRIPTION\":\"AMBASSADORS COURT\",\"TOWN_NAME\":\"LONDON\",\"ADMINISTRATIVE_AREA\":\"CITY OF WESTMINSTER\",\"POSTCODE_LOCATOR\":\"SW1A 1AA\"}"
}
```

So, we need to dig out the postcode from either a string (`:text`) or worse a JSON object (`:value`).  If I remember correctly, this is the main reason why we put the postcode into the session.

leenagupte (approved):

PR #410, opened 2020-06-12
Redact IDs from urls for GA

Rosa-Fox:
What
----

This will redact any value in the query string (`window.location.search`) parameters, but keep the key present - this ensures that Google Analytics knows about the keys, but not the values of those keys. This will prevent any personally identifiable information (PII) from leaking into GA.

For example:

`https://example.com/?really_important_thing=top_secret` 

will be reported as 

`https://example.com/?really_important_thing=<REALLY_IMPORTANT_THING>`

We remove the `?` from `window.location.search`, and then `split()` that by the ampersands. We then split the `key=value` into a key and value. 

The query string is then put together without the value - `key=<KEY>` - and the whole URL is constructed again to be returned as a string for reporting into Google Analytics.

Jasmine tests have been added as part of this pull request - `bundle exec rake jasmine:ci` to run them, or `bundle exec rake` to run all tests.

## Why

Because PII in GA is bad and this will help prevent any PII from being reported.

How to review
-------------

- The Google doc linked to the Trello card may also be useful for giving some more context
- Check the Jasmine tests cover all bases

Links
-----

[Trello](https://trello.com/c/J5W2oQmv/569-dont-send-google-analytics-user-transaction-id-query-parameters-for-vulnerable-people-form)


app/assets/javascripts/analytics.js.erb
```diff
@@ -1,7 +1,21 @@
...
+var stripPII = function(url) {
+  url = url.replace(REFERENCE_NUMBER_PATTERN, 'reference_number')
+  url = url.replace('true', 'contact_gp')
+  url = url.replace('false', 'contact_gp')
+
+  if(url.includes('reference_number') || 'contact_gp') {
+    return url
+  } else {
+    return location.protocol + '//' + location.host + location.pathname
+  }
```
leenagupte: I don't think I understand why this `if/else` is needed. What happens if the function always returns `url`?

Rosa-Fox: This was a result of the conversation [here](https://github.com/alphagov/govuk-coronavirus-business-volunteer-form/pull/361). Originally I just returned `url` but using `if/else` was cautionary incase other params are added and we miss changing this code. Have realised its still a bit flawed in that if only one of `reference_number` or `contact_gp` are matched then the other is not redacted, will change to an `&&`...

leenagupte: I didn't realise that we're trying to avoid sending other params that could be in the query-string to GA too.  Do you think it's possible to loop through all/any of the query params and replace them all with "XXXXXXX"? It means that if `contact_gp` for example is ever renamed, we don't need to worry about having to update it here too.

leenagupte: I've fallen down a rabbit hole. Apparently there's a `URLSearchParams` class that gets query params and interrogates query params: https://flaviocopes.com/urlsearchparams/

Rosa-Fox: nice! Thanks, I will try that :)

Rosa-Fox: @leenagupte thanks for suggested, is a much stronger approach. I have amended the commit.

app/assets/javascripts/analytics.js.erb
```diff
@@ -2,6 +2,14 @@
 var element = document.querySelector('[data-module="cookie-banner"]')
 new GOVUK.Modules.CookieBanner().start($(element))
 
+var stripPII = function() {
+  const params = new URLSearchParams(window.location.search)
+  for (const [key, value] of params.entries()) {
+    params.set(key, "XXXX")
+  }
+  return location.protocol + '//' + location.host + location.pathname + params
```
leenagupte:
```suggestion
  return location.protocol + '//' + location.host + location.pathname + '?' + params
```

Without the '?' the url returned looks like: 
```
"http://localhost:5000/confirmationcontact_gp=XXXX&reference_number=XXXX"
```

I'm not sure if it's ok to add a trailing `?` to urls if there aren't any querystring params. It might be worth returning `location.protocol + '//' + location.host + location.pathname` if params is `null` before doing the loop?

leenagupte (commented):
This is looking good. I'm excited that `URLSearchParams` works!

I've spotted one small thing in the url generated, otherwise it looks good.

I think you definitely should get someone how knows Javascript to look this over as well though.

app/assets/javascripts/analytics.js.erb
Rosa-Fox: ahh! have amended :) @leenagupte

leenagupte (dismissed):
💯 👍 🎉

app/assets/javascripts/analytics.js.erb
```diff
@@ -2,6 +2,25 @@
 var element = document.querySelector('[data-module="cookie-banner"]')
 new GOVUK.Modules.CookieBanner().start($(element))
 
+var redactPII = function() {
+  const params = new URLSearchParams(window.location.search)
```
alex-ju: `URLSearchParams` is not supported in Internet Explorer (11 and below) so we can't rely on it, unfortunately.

app/assets/javascripts/analytics.js.erb
```diff
@@ -2,6 +2,25 @@
...
+  for (const [key, value] of params.entries()) {
```
alex-ju: We don't use `const` and `let`, only `var` as we don't transpile from ES6 and the new declarations are not supported in Internet Explorer (in v11 not supported in for-in and for-of loops and below v11 there is no support whatsoever).

alex-ju (commented):
Looks good, but unfortunately this approach doesn't cover all the supported browsers so if we must do it, it will have to be regex. I would be cautious about adding complexity to our scripts (that are served to every visitor) and given this isn't technically PII I would suggest we exclude them on the GA-side using 'Exclude Query Parameters'.

alex-ju (approved):

Rosa-Fox:
This looks great, thanks so much @injms! Thanks for taking me through the code, I have tested the updated changes in the browser and run locally... but I can't approve as I originally opened them PR.

injms:
> This looks great, thanks so much @injms! Thanks for taking me through the code, I have tested the updated changes in the browser and run locally... but I can't approve as I originally opened them PR.

Ah, okay - makes sense. @alex-ju, could you review this again please? 🙏

alex-ju (approved):
Re-checked. It's aligned with https://github.com/alphagov/govuk-coronavirus-business-volunteer-form/pull/389, happy to get it in 👏

PR #411, opened 2020-06-12
Bug - possible return of nil

gclssvglx:

Previous work had sought to resolve these issues.  However, it has been [noted](https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/pull/406#discussion_r439323912) (thanks @leenagupte) that a `nil` return is still possible as calling presence after the strip_tags method on a string containing only tags will return nil.

This change fixes that issue and moves two methods into the AddressHelper, so that they can tested more easily, and refactors the moved methods to dry up the code.

Links
-----

[Incident Google Doc](https://docs.google.com/document/d/16qU8ZgSI7Y_IWN7F1nCkLWtVFvX_jw8VWwAiT9rs7EI/edit#)
[Incident Investigation Trello Card](https://trello.com/c/U8FR01EG/581-investigate-root-cause-of-incident)

### DIFF
--- a/app/controllers/coronavirus_form/support_address_controller.rb
+++ b/app/controllers/coronavirus_form/support_address_controller.rb
@@ -56,20 +56,12 @@ private
 
   def given_address
     {
-      building_and_street_line_1: strip_tags_from_address_field(params[:building_and_street_line_1]),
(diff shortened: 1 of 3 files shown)


leenagupte (approved):
👍 Good call on moving this to a helper and adding tests.

PR #412, opened 2020-06-12
RFC for question routing defined in config.

steveholmes:
Links
-----

https://trello.com/c/p7dDrBKK/539-figure-out-how-routing-in-cev-app-can-be-defined-as-config-not-code

https://trello.com/c/LNk85mPk/26-represent-questions-as-data

### DIFF
--- a/docs/rfc/rfc-vpf-001-question-routing.md
+++ b/docs/rfc/rfc-vpf-001-question-routing.md
@@ -1,0 +1,97 @@
+# Define routing between pages using config, not code 
+
+## Summary
+
+We have various controllers in the app that look like:
+
+```ruby
+  def next_page_url
+    return dietary_requirements_url if answer_next_question?
+    return check_your_answers_url if session[:check_answers_seen]
+
+    basic_care_needs_url
+  end
+```
+
+We want anyone to easily change the routing of questions, and the conditional logic that determines which questions a user sees next.
+This is an important but complex part of making the apps more data driven.
+
+## Problem
+
+At the moment all questions are coded as a separate controller. This makes adding new questions or changing the order quite cumbersome.
+Secondly, if we deploy changes to the order then current users could end up completing their journey without answering all questions.
+We would like to solve these by using configuration files (which in the future could be created by a publishing app) to define the flow.  
+
+## Proposal
+
+It is proposed that a YAML file can be used to define the flow through the forms. A single controller would handle the pages that are defined in the YAML file.
+It is proposed that an Interactor pattern is used to handle the validation (beyond the scope of this RFC) and return the next path (assuming validation passes)
+ [Interactors in Ruby](https://goiabada.blog/interactors-in-ruby-easy-as-cake-simple-as-pie-33f66de2eb78)
+
+### Simple flows
+Many of the forms have a simple flow which involves checking for invalid values on the form and then progressing to the next page if no validation problems are found.
+(The validation is beyond the scope of this RFC) 
+An example of this would be the `name` controller which checks for validation and then proceeds to the `date_of_birth` controller. This type of progress could simply be defined in YAML as
+```yaml
+name:
+  default_next_question: 
+    question_path: date_of_birth
+``` 
+
+(`question_path` would signify the next path handled by the YAML file and `path` could be used to signify a request to a normal rails route)
+
+An alternative could be that rather than defining the default next question, that could be worked out from the order of the questions in the YAML file? 
+We already do this to determine the order that the questions are played back on the check answers page. Perhaps instead this could define the "not-happy" path?
+But if people are manually copy/pasting a question to a different section of the yaml file then as the flow isn't explicit and requires a bit more thinking - possibly causing manual error.
+
+### Simple conditionals
+The next simplest definition would be around a different path being taken depending upon an answer selected and stored
+For example, if the user does not live in england we direct them to `not_eligible_england` controller 
+```yaml
+live_in_england:
+  default_next_question: 
+    question_path: nhs_letter
+  conditional_paths:
+  - conditions:
+      live_in_england:
+      - 'No'
+      question_path: not_eligible_england
+```
+Note: We should use the option key here rather than the value in the content as that is more likely to change. But it is left as a literal at the moment for simplicity of the exmaple
+
+More complex conditionals could be combined using `and` and `or` logic, similar to [JsonLogic](http://jsonlogic.com/) but is currently
+not proposed (see Custom Routing) and instead is something to consider as we further develop (YAGNI).
+
+### Check answers seen
+A further complexity for the routing is that the user may return to a page/section from the "Check Answers" page. 
+Any implementation must be able to recognise this and return to the check answers page after a section is complete.
+The proposal is that each path can be marked with a flag for `redirect_if_check_answers_seen: true` which will move back to the check answers page as it currently does
+```yaml
+nhs_letter:
+  default_next_question:
+    redirect_if_check_answers_seen: false
+    question_path: name
+...
+```
+As most questions will route back to the check answers page, this could be the default behaviour and only need to add `redirect_if_check_answers_seen: false` for those cases and true would be assumed for others.
+
+### Custom routing
+The most complex path routing happens when the conditional is defined using ruby statements and expressions.
+The two examples of this are (i) in the `postcode_lookup` controller and (ii) `essential_supplies` controller.
+At the moment it is recommended that these are marked in the YAML as "custom". When the handler (`interactor`?) encounters a `custom`
+conditional, then it would look for a custom method in the app that could define the next path. 
+These custom routing methods could also handle the possible circular routes, for example, repeating a path for multiple supplies in the
+business support app.
+Assuming that the name of the interactor matches the name of the question, a custom property may not need to be defined 
+as it's possible to use the question name to determine if the interactor class exists (has been defined).
+
+### Mandatory questions and deploying flow changes
+One of the problems with the current app is that changing the flow can affect users part way through there journey.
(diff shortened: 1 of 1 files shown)


docs/rfc/rfc-vpf-001-question-routing.md
```diff
@@ -0,0 +1,88 @@
...
+## Problem
+
+At the moment all questions are coded as a separate controller. This makes adding new questions or changing the order quite cumbersome.
+Secondly, if we deploy changes to the order then current users could end up completing their journey without answering all questions.
+We would like to solve these by using configuration files (which in the future could be created by a publishing app) to define the flow.  
+
+## Proposal
+
+It is proposed that a YAML file can be used to define the flow through the forms. A single controller would handle the pages that are defined in the YAML file.
+It is proposed that an Interactor pattern is used to handle the validation (beyond the scope of this RFC) and return the next path (assuming validation passes)
```
leenagupte: We use the Interactor in Content Publisher, see: https://github.com/alphagov/content-publisher/blob/master/docs/adr/0011-moving-business-logic-out-of-controllers.md, but for those unfamiliar with the pattern, it's worth linking to some docs for context.

docs/rfc/rfc-vpf-001-question-routing.md
```diff
@@ -0,0 +1,88 @@
...
+It is proposed that a YAML file can be used to define the flow through the forms. A single controller would handle the pages that are defined in the YAML file.
+It is proposed that an Interactor pattern is used to handle the validation (beyond the scope of this RFC) and return the next path (assuming validation passes) 
+
+### Simple flows
+Many of the forms have a simple flow which involves checking for invalid values on the form and then progressing to the next page if no validation problems are found.
+(The validation is beyond the scope of this RFC) 
+An example of this would be the `name` controller which checks for validation and then proceeds to the `date_of_birth` controller. This type of progress could simply be defined in YAML as
+```yaml
+name:
+  default_next_question:
```
leenagupte, line 37: Rather than defining the default next question, could that be worked out from the order of the questions in the YAML file? We already do this to determine the order that the questions are played back on the check answers page. Perhaps instead this could define the "not-happy" path?

docs/rfc/rfc-vpf-001-question-routing.md
```diff
@@ -0,0 +1,88 @@
...
+Many of the forms have a simple flow which involves checking for invalid values on the form and then progressing to the next page if no validation problems are found.
+(The validation is beyond the scope of this RFC) 
+An example of this would be the `name` controller which checks for validation and then proceeds to the `date_of_birth` controller. This type of progress could simply be defined in YAML as
+```yaml
+name:
+  default_next_question: 
+    question_path: date_of_birth
+``` 
+
+(`question_path` would signify the next path handled by the YAML file and `path` could be used to signify a request to a normal rails route)
```
leenagupte, line 41: Does `path` refer to the path helper method here?

docs/rfc/rfc-vpf-001-question-routing.md
```diff
@@ -0,0 +1,88 @@
...
+
+### Simple conditionals
+The next simplest definition would be around a different path being taken depending upon an answer selected and stored
+For example, if the user does not live in england we direct them to `not_eligible_england` controller 
+```yaml
+live_in_england:
+  default_next_question: 
+    question_path: nhs_letter
+  conditional_paths:
```
leenagupte, line 54: This is what I meant above! The conditional path!

docs/rfc/rfc-vpf-001-question-routing.md
```diff
@@ -0,0 +1,88 @@
...
+  - conditions:
+      live_in_england:
+      - 'No'
```
leenagupte, line 57: Nitpick: We should perhaps use the option key here rather than the value in the content as that is more likely to change.

docs/rfc/rfc-vpf-001-question-routing.md
```diff
@@ -0,0 +1,88 @@
...
+      question_path: not_eligible_england
+```
+
+More complex conditionals could be combined using `and` and `or` logic, similar to [JsonLogic](http://jsonlogic.com/) but is currently
+not proposed (see Custom Routing) and instead is something to consider as we further develop (YAGNI).
+
+### Check answers seen
+A further complexity for the routing is that the user may return to a page/section from the "Check Answers" page. 
+Any implementation must be able to recognise this and return to the check answers page after a section is complete.
+The proposal is that each path can be marked with a flag for `redirect_if_check_answers_seen: true` which will move back to the check answers page as it currently does
```
leenagupte, line 68: I think there are only a handful of questions that don't route users straight back to the check answers page, so perhaps the exception could be defined in the yaml, and default behaviour assumed for the rest?

docs/rfc/rfc-vpf-001-question-routing.md
```diff
@@ -0,0 +1,88 @@
...
+  default_next_question:
+    redirect_if_check_answers_seen: true
+    question_path: name
+...
+```
+
+### Custom routing
+The most complex path routing happens when the conditional is defined using ruby statements and expressions.
+The two examples of this are (i) in the `postcode_lookup` controller and (ii) `essential_supplies` controller.
+At the moment it is recommended that these are marked in the YAML as "custom". When the handler (`interactor`?) encounters a `custom`
```
leenagupte, line 81: Assuming that the name of the interactor matches the name of the question, I'm not sure if a `custom` property needs to be defined as I think it's possible to use the question name to determine if the interactor class exists (has been defined).

docs/rfc/rfc-vpf-001-question-routing.md
```diff
@@ -0,0 +1,88 @@
...
+These custom routing methods could also handle the possible circular routes, for example, repeating a path for multiple supplies in the
+business support app.
+
+### Mandatory questions and deploying flow changes
+One of the problems with the current app is that changing the flow can affect users part way through there journey.
+By defining the paths in this file, along with the conditions for branching, the app should be able to check the "path journey" on each
+request and ensure that the path can be traced up to the current request. Any missing questions could cause the request to be redirected to the missing question.
+
+Logic for determining any missed question could be:
+1. The controller has a list of all questions fro the YAML file, that it must prune to find the next question
```
leenagupte:
```suggestion
1. The controller has a list of all questions from the YAML file, that it must prune to find the next question
```

leenagupte (commented):
The overall approach sounds good to me. I especially like the idea of using an interactor to do the routing.

I've add a few of thoughts/comments inline. Happy to discuss more in person.

docs/rfc/rfc-vpf-001-question-routing.md
```diff
@@ -25,7 +25,7 @@ We would like to solve these by using configuration files (which in the future c
 ## Proposal
 
 It is proposed that a YAML file can be used to define the flow through the forms. A single controller would handle the pages that are defined in the YAML file.
-It is proposed that an Interactor pattern is used to handle the validation (beyond the scope of this RFC) and return the next path (assuming validation passes) 
+It is proposed that an Interactor pattern is used to handle the validation (beyond the scope of this RFC) and return the next path (assuming validation passes) [https://goiabada.blog/interactors-in-ruby-easy-as-cake-simple-as-pie-33f66de2eb78](Interactors in Ruby)
```
leenagupte:
```suggestion
It is proposed that an Interactor pattern is used to handle the validation (beyond the scope of this RFC) and return the next path (assuming validation passes) [Interactors in Ruby](https://goiabada.blog/interactors-in-ruby-easy-as-cake-simple-as-pie-33f66de2eb78)
```

The link is broken.

docs/rfc/rfc-vpf-001-question-routing.md
steveholmes: My only worry for having a default would be if people are manually copy/pasting a question to a different section of the yaml file then as the flow isn't explicit and requires a bit more thinking - possibly causing manual error. But add the alternative to the file.

docs/rfc/rfc-vpf-001-question-routing.md
steveholmes: yes that's right.

docs/rfc/rfc-vpf-001-question-routing.md
steveholmes: doh! Thank you

Rosa-Fox:
We had a team discussion and decided that we will try spiking this approach on the EV application: https://trello.com/c/G2DRC9t7/319-routing-as-config-spike

Rosa-Fox (approved):

reggieb:
Have you considered using [form objects](https://thoughtbot.com/blog/activemodel-form-objects)? They have a similar pattern to Interactors but also add some of the niceness `ActiveModel` provides: especially validation. I've used form object combined with state machines to create multi-step forms like these with a lot of success in the past. A state machine can be over-kill if the path through the form is linear, but can make multi-pathway forms a lot cleaner.

As an example of form object usage, we used them in the MoJ Apply for legal aid app: https://github.com/ministryofjustice/laa-apply-for-legal-aid/tree/master/app/forms

reggieb:
Another advantage of using ActiveModel form objects is that they behave in the way rails form elements expect objects to behave. So a lot of the form boiler plate code can be removed.

reggieb:
On the Apply for legal aid app, we used a set of files to define the routes in a similar way to that being proposed. If memory servers me right, we started with YAML files, but then moved to hashes because some of the decisions involved assessing dynamic content (if x route to y else route to z) which was easier to configure within hashes. For an example:

https://github.com/ministryofjustice/laa-apply-for-legal-aid/blob/master/app/services/flow/flows/citizen_start.rb

PR #439, opened 2020-07-02
Add feature test for minimal information

GDSNewt:
What
----

This feature test will check that a user can complete the form journey
providing the bare minimum information required.

I have also removed the abilty for nhs_number to be null as it is
now always required.

Why
----
We had an incident in which it was discovered that a lot of the errors were being caused by users not completing some fields. In all cases, these were fields that is user is allowed to leave blank, for example, `building_and_street_line_2` in the support address controller.

These values aren't just used in the controller where they're first stored, but across the app. So we need to check that these values can be `nil` at various places in the journey.

[Trello card](https://trello.com/c/dxNWr68o/615-add-a-feature-test-to-check-that-users-can-submit-forms-with-the-minimum-amount-of-information)

### DIFF
--- a/config/schemas/form_response.json
+++ b/config/schemas/form_response.json
@@ -131,7 +131,7 @@
       ]
     },
     "nhs_number": {
-      "type": ["string", "null"],
+      "type": "string",
       "pattern": "^[0-9]*$"
     },
     "essential_supplies": {
--- a/spec/features/filling_in_the_form/fill_in_the_form_spec.rb
+++ b/spec/features/filling_in_the_form/fill_in_the_form_spec.rb
@@ -15,7 +15,7 @@ RSpec.feature "fill in the vulnerable people form" do
       and_has_given_their_date_of_birth
       and_has_given_their_postcode
       and_has_selected_their_address
-      and_has_submited_their_address
+      and_has_submitted_their_address
       and_has_given_their_contact_details
       and_has_checked_their_contact_details
       and_is_not_getting_any_essential_supplies
@@ -25,6 +25,25 @@ RSpec.feature "fill in the vulnerable people form" do
       and_has_accepted_the_terms_and_conditions
       then_they_can_be_supported
     end
+  end
+
+  scenario "fill in the form with minimal information" do
+    given_an_extremely_vulnerable_person_during_the_covid_19_pandemic
+    that_lives_in_england
+    and_has_recently_received_an_nhs_letter
+    and_has_given_their_nhs_number
+    and_has_given_the_minimal_name_information
+    and_has_given_their_date_of_birth
+    and_has_given_their_postcode
+    and_has_selected_their_address
+    and_has_submitted_the_miminum_address_information
+    and_has_given_no_contact_details
+    and_is_not_getting_any_essential_supplies
+    and_does_not_have_any_special_dietary_requirements
+    where_there_is_no_one_available_to_carry_supplies
+    and_their_basic_care_needs_are_not_being_met
+    and_has_accepted_the_terms_and_conditions
+    then_they_can_be_supported
   end
 
   describe "fill in the form" do
--- a/spec/support/fill_in_the_form_steps.rb
+++ b/spec/support/fill_in_the_form_steps.rb
@@ -63,6 +63,15 @@ module FillInTheFormSteps
     end
   end
 
+  def and_has_given_the_minimal_name_information
+    expect(page).to have_content(I18n.t("coronavirus_form.questions.name.title"))
+    within find(".govuk-main-wrapper") do
+      fill_in "first_name", with: "Test Please Ignore"
+      fill_in "last_name", with: "Test Please Ignore"
+      click_on I18n.t("coronavirus_form.submit_and_next")
+    end
+  end
+
   def and_has_given_their_date_of_birth
     expect(page).to have_content(I18n.t("coronavirus_form.questions.date_of_birth.title"))
     within find(".govuk-main-wrapper") do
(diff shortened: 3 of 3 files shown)


spec/support/fill_in_the_form_steps.rb
```diff
@@ -106,6 +116,18 @@ def and_has_submited_their_address
     end
   end
 
+  def and_has_submited_the_miminum_address_information
```
leenagupte:
```suggestion
  def and_has_submitted_the_minimum_address_information
```

spec/support/fill_in_the_form_steps.rb
```diff
@@ -106,6 +116,18 @@ def and_has_submited_their_address
...
+    expect(page).to have_content(I18n.t("coronavirus_form.questions.support_address.title"))
+    within find(".govuk-main-wrapper") do
+      fill_in "building_and_street_line_1", with: "Test Please Ignore"
+      fill_in "building_and_street_line_2", with: ""
+      fill_in "town_city", with: "Test Please Ignore"
+      fill_in "county", with: ""
```
leenagupte: Can this line be removed from the test? It feels odd "filling in" something that the user hasn't touched.

spec/support/fill_in_the_form_steps.rb
```diff
@@ -116,6 +138,16 @@ def and_has_given_their_contact_details
     end
   end
 
+  def and_has_given_no_contact_details
+    expect(page).to have_content(I18n.t("coronavirus_form.questions.contact_details.title"))
+    within find(".govuk-main-wrapper") do
+      fill_in "phone_number_calls", with: ""
+      fill_in "phone_number_texts", with: ""
+      fill_in "email", with: ""
```
leenagupte: Can these lines be removed? We want to test users only clicking the start button if possible.

config/schemas/form_response.json
```diff
@@ -131,7 +131,7 @@
       ]
     },
     "nhs_number": {
-      "type": ["string", "null"],
+      "type": "string",
```
leenagupte, line 134: Good spot!

spec/features/filling_in_the_form/fill_in_the_form_spec.rb
```diff
@@ -27,6 +27,25 @@
...
+  scenario "fill in the form with minimal information" do
+    given_an_extremely_vulnerable_person_during_the_covid_19_pandemic
+    that_lives_in_england
+    and_has_recently_received_an_nhs_letter
+    and_has_given_their_nhs_number
+    and_has_given_the_minimal_name_information
+    and_has_given_their_date_of_birth
+    and_has_given_their_postcode
+    and_has_selected_their_address
+    and_has_submited_the_miminum_address_information
```
leenagupte:
```suggestion
    and_has_submitted_the_minimum_address_information
```

spec/support/fill_in_the_form_steps.rb
```diff
@@ -63,6 +63,16 @@ def and_has_given_their_name
     end
   end
 
+  def and_has_given_the_minimal_name_information
+    expect(page).to have_content(I18n.t("coronavirus_form.questions.name.title"))
+    within find(".govuk-main-wrapper") do
+      fill_in "first_name", with: "Test Please Ignore"
+      fill_in "middle_name", with: ""
```
leenagupte: Can this line be removed from the test? It feels odd "filling in" something that the user hasn't touched.

leenagupte (commented):
This is looking good.

I've added a couple of inline comments, but nothing major.

spec/support/fill_in_the_form_steps.rb
GDSNewt: I believe the empty string 'fill in' on line 123 has to be there to make the test in any way meaningful. I'm not sure exactly how it works but I suspect the validation doesn't perform any checks if you omit the 'fill_in' for a field. For example line 122, which is definitely required, If i remove that line the tests still pass whereas if I give it an empty string it fails. 

I believe this is true for the other lines with empty strings that you highlighted.

spec/support/fill_in_the_form_steps.rb
reggieb: That's quite easy to test. I added `p params` as line 5 of `/app/controllers/coronavirus_form/name_controller.rb` and then ran `rspec  spec/features/filling_in_the_form/fill_in_the_form_spec.rb:30` with the code as is. The `params` output was:
```ruby
<ActionController::Parameters {"first_name"=>"Test Please Ignore", "middle_name"=>"", "last_name"=>"Test Please Ignore", "controller"=>"coronavirus_form/name", "action"=>"submit"} permitted: false>
```
Then I commented out line 70 and the output looked like this:
```ruby
<ActionController::Parameters {"first_name"=>"Test Please Ignore", "middle_name"=>"", "last_name"=>"Test Please Ignore", "controller"=>"coronavirus_form/name", "action"=>"submit"} permitted: false>
```
As you can see, exactly the same.

These tests are using capybara, so the actions are being carried out by a headless browser - therefore the page will be rendered headlessly, and the submit button clicked. If nothing is entered in a text field it will behave the same as with a normal browser - an empty string is submitted.

So - long winded - but in conclusion these `fill_in "<xxxx>", with ""` are redundant.

spec/support/fill_in_the_form_steps.rb
```diff
@@ -63,6 +63,16 @@ def and_has_given_their_name
     end
   end
 
+  def and_has_given_the_minimal_name_information
+    expect(page).to have_content(I18n.t("coronavirus_form.questions.name.title"))
+    within find(".govuk-main-wrapper") do
+      fill_in "first_name", with: "Test Please Ignore"
```
Rosa-Fox, line 69: We can leave the text as it is, but just incase you didn't know, we don't write any of the smoke test inputs to the database thanks to [this line](https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/blob/9fd651d30f9cd06074bcedfb4cf0e9f21b80d9d8/spec/spec_helper.rb#L25).

leenagupte (approved):
🎉 LGTM! 👍

PR #442, opened 2020-07-08
644 Reduce times session is duplicated as a hash in check answers controller

reggieb:
What
----
An Intermittent bug appears to be caused by either 

- errors in the [**json-schema** gem](https://github.com/ruby-json-schema/json-schema). The main JSON::Validator class has lots of class variables, deeply nested methods (e.g the build_schemas method is 65 lines), and a lot of the processing is done at the class level - so against the class singleton (34 methods within the class << self declaration). I expect that's why they're using Mutex to try and tame threading issues. This PR replaces json-schema with the [**json_schemer** gem](https://github.com/davishmcclurg/json_schemer) which has a simpler structure and less reliance of processing within class singletons.
- missing keys in a hash (derived from session) on being passed to JSON schema validator. The changes in this PR reduce the manipulation of session in case this is the cause of the error.

The bug generates errors like: 
```
[ The property '#/' did not contain a required property of 'name' in schema b48faf71-f8f7-5ab7-9536-fbf40922a7fa , The property '#/' did not contain a required property of 'know_nhs_number' in schema b48faf71-f8f7-5ab7-9536-fbf40922a7fa ]
```
This error can be replicated if keys are removed from the hash passed to the JSON validator.

Prior to this change, `session` was being manipulated by being converted into a hash and then converted to a `ActiveSupport::HashWithIndifferentAccess` (I believe each change creates a duplicate object in memory). As `session` already has indifferent access (keys can be identified by symbol or string versions of the keys) these changes are unnecessary manipulation.

The one place where manipulation is needed is when `except` needs to be called on the session hash (which doesn't respond to `except`). Here, as long a string keys are used in the `except` clause, just a single `to_h` manipulation is needed.

Links
-----

- [Trello card](https://trello.com/c/5BDly30W/644-investigate-formresponseinvaliderror-in-sentry)
- [Errors in sentry](https://sentry.io/organizations/govuk/issues/1667664108/?project=5170680)


steveholmes (approved):
Yes, hard to tell what will fix the problem as we can't reproduce locally, but looks good to me :+1:

Gemfile
```diff
@@ -14,7 +14,7 @@ gem "faraday", "~> 1.0.1"
 gem "fog-aws"
 gem "govuk_app_config", "~> 2.2.0"
 gem "govuk_publishing_components", "~> 21.57.1"
-gem "json-schema", "~> 2.8.1"
+gem "json_schemer", "~> 0.2"
```
leenagupte, line 17: @reggieb Can you explain why you're using this gem in the commit message. i.e. What's the benefit of this change? Also, can you add a link to the gem documentation.

spec/helpers/schema_helper_spec.rb
```diff
@@ -4,332 +4,26 @@
...
-        )
-        expect(validate_against_form_response_schema(data).first).to include(
-          "check_answers_seen",
-          "_csrf_token",
-          "current_path",
-          "previous_path",
-          "session_id",
-        )
+      it "returns a fail as true" do
+        expect(subject.fail).to be(true)
```
leenagupte: I don't understand what this is testing. What does "fail" mean?

leenagupte (dismissed):
@reggieb I'm concerned by the changes made in the second commit. Most especially the removal of the tests used to make sure that the schema validates as expected. 

None of us are experts at writing schemas, so we used these tests to make sure that we are writing the schema correctly.

For example, we used the tests to make sure that we aren't doing things like making a value required when it isn't supposed to be, or that we aren't allowing users to just save any value in a field.

Also, we had plans to enforce the json validation at the model level, and these tests were added to give us confidence that we wouldn't be preventing users from submitting the form because the schema wasn't quite right. The model validation was delayed due to the occurrences of the error you're trying to fix in the first commit, but it's possible that it _may_ be added sometime in the future.

Also, as mentioned in the inline comments, can you explain why the json_schemer is being used. What is the benefit of it over using the existing json-schema gem?

Happy to talk this through in person.

spec/helpers/schema_helper_spec.rb
reggieb: I'll rename it `invalid`

Gemfile
reggieb: I've updated the commit message

spec/helpers/schema_helper_spec.rb
```diff
@@ -4,332 +4,26 @@
...
-      it "allows medical_conditions to be blank" do
-        data = valid_data.except(:medical_conditions)
-        expect(validate_against_form_response_schema(data)).to be_empty
-      end
-
-      it "returns a list of errors when medical_conditions has an unexpected value" do
-        data = valid_data.merge(medical_conditions: "Foo")
-        expect(validate_against_form_response_schema(data).first).to include("medical_conditions")
-      end
+    it "returns no error if the data is valid" do
```
leenagupte:
```suggestion
    it "returns no errors if the data is valid" do
```

spec/helpers/schema_helper_spec.rb
leenagupte: I'm still not sure what this is testing.

Could you also add a test for if there are no validation errors? For example, what does `validate_against_form_response_schema` return if the data validates against the schema?

spec/helpers/schema_helper_spec.rb
```diff
@@ -4,332 +4,26 @@
...
-    describe "other information" do
-      it "allows a reference_id to be stored" do
-        data = valid_data.merge(reference_id: "abc")
-        expect(validate_against_form_response_schema(data)).to be_empty
-      end
-
-      it "returns a list of errors when check_answers_seen has an invalid value" do
-        data = valid_data.merge(check_answers_seen: "Foo")
-        expect(validate_against_form_response_schema(data).first).to include("check_answers_seen")
+      it "returns a list of errors when live_in_england is missing" do
```
leenagupte: It looks all of the schema validation tests have been moved into a service in the next commit. Should this test join them there?

app/controllers/coronavirus_form/check_answers_controller.rb
```diff
@@ -16,7 +16,7 @@ def submit
     session[:reference_id] = submission_reference
 
     json_validation = validate_against_form_response_schema(sanitised_session)
-    if json_validation.fail
+    if json_validation.invalid?
```
leenagupte:
```suggestion
    unless json_validation.valid?
```
Generally `valid?` is called to test if something is ok. It would be nice if we could repeat that pattern here.

Can this change in name be fixed up to the previous commit to make it easier to traverse the history.

app/helpers/schema_helper.rb
```diff
@@ -2,14 +2,7 @@ module SchemaHelper
...
-    data = data.deep_stringify_keys
     schema = JSON.parse(FORM_RESPONSE_SCHEMA)
-    validator = JSONSchemer.schema(schema)
-    errors = validator.validate(data).to_a
-
-    OpenStruct.new(
-      fail: errors.any?,
-      errors: errors.map { |report| report["details"] },
-    )
+    JsonValidator.new schema, data
```
leenagupte:
```suggestion
    JsonValidator.new(schema, data)
```

This is much nicer to read!

app/services/json_validator.rb
```diff
@@ -0,0 +1,38 @@
...
+  attr_reader :schema, :data
+
+  def initialize(schema, data)
+    @schema = schema
+    @data = data.deep_stringify_keys
+  end
+
+  def invalid?
+    !validator.valid?(data)
+  end
```
leenagupte:
```suggestion
  def valid?
    validator.valid?(data)
  end
```

I find truthy statements are easier to read.

spec/services/json_validator_spec.rb
```diff
@@ -0,0 +1,517 @@
...
+    it "returns false if data is valid" do
+      expect(subject.invalid?).to be(false)
+    end
+
+    context "with invalid data" do
+      let(:data) { valid_data.except(:live_in_england) }
+
+      it "returns true" do
+        expect(subject.invalid?).to be(true)
+      end
```
leenagupte: This looks like it's a duplicate of the "live in england" test below, so can be removed.

spec/services/json_validator_spec.rb
```diff
@@ -0,0 +1,517 @@
...
+
+  let(:data) { valid_data }
+  let(:schema_data) { File.read(Rails.root.join("config/schemas/form_response.json")) }
+  let(:schema) { JSON.parse schema_data }
+  subject { described_class.new(schema, data) }
+
+  describe "#invalid?" do
+    it "returns false if data is valid" do
+      expect(subject.invalid?).to be(false)
+    end
```
leenagupte: As stated before, I think this would be easier to read if the method was testing that the data is valid rather than invalid.

spec/services/json_validator_spec.rb
```diff
@@ -0,0 +1,517 @@
...
+    context "when live_in_england missing" do
+      let(:data) { valid_data.except(:live_in_england) }
+
+      it "returns a missing keys error" do
+        expect(error["missing_keys"]).to contain_exactly("live_in_england")
+        expect(error["data"]).to be_nil
+      end
+    end
+
+    context "when live_in_england in valid" do
```
leenagupte:
```suggestion
    context "when live_in_england is invalid" do
```

spec/services/json_validator_spec.rb
```diff
@@ -0,0 +1,517 @@
...
+      let(:data) { valid_data.merge(live_in_england: "invalid") }
+
+      it "returns a yes_no definition error against field" do
```
leenagupte:
```suggestion
      it "returns a yes_no definition error against the field" do
```

spec/services/json_validator_spec.rb
```diff
@@ -0,0 +1,517 @@
...
+  end
+
+  describe "#errors" do
+    let(:error) { subject.errors.first }
+
+    it "returns no error if the data is valid" do
+      expect(subject.errors).to be_empty
+    end
+
+    context "when live_in_england missing" do
```
leenagupte:
```suggestion
    context "when live_in_england is missing" do
```

spec/services/json_validator_spec.rb
```diff
@@ -0,0 +1,517 @@
...
+      let(:data) { valid_data.merge(live_in_england: "invalid") }
+
+      it "returns a yes_no definition error against field" do
+        expect(error["data"]).to eq("invalid")
+        expect(error["data_pointer"]).to include("live_in_england")
+        expect(error["schema_pointer"]).to eq("/definitions/yes_no")
+      end
+    end
+
+    context "when nhs_letter missing" do
```
leenagupte:
```suggestion
    context "when nhs_letter is missing" do
```

spec/services/json_validator_spec.rb
```diff
@@ -0,0 +1,517 @@
...
+      let(:data) { valid_data.except(:nhs_letter) }
+
+      it "returns a missing keys error" do
+        expect(error["missing_keys"]).to contain_exactly("nhs_letter")
+      end
+    end
+
+    context "when nhs_letter invalid" do
```
leenagupte:
```suggestion
    context "when nhs_letter is invalid" do
```

spec/services/json_validator_spec.rb
```diff
@@ -0,0 +1,517 @@
...
+      let(:data) { valid_data.merge(nhs_letter: "invalid") }
+
+      it "returns a nhs_letter property error" do
```
leenagupte:
```suggestion
      it "returns an nhs_letter property error" do
```

spec/services/json_validator_spec.rb
```diff
@@ -0,0 +1,517 @@
...
+        expect(error["data"]).to eq("invalid")
+        expect(error["data_pointer"]).to include("nhs_letter")
+        expect(error["schema_pointer"]).to eq("/properties/nhs_letter")
+      end
+    end
+
+    context "when medical_conditions blank" do
```
leenagupte:
```suggestion
    context "when medical_conditions is blank" do
```

spec/services/json_validator_spec.rb
```diff
@@ -0,0 +1,517 @@
...
+      let(:data) { valid_data.except(:medical_conditions) }
+
+      it "returns no error" do
```
leenagupte:
```suggestion
      it "returns no errors" do
```

spec/services/json_validator_spec.rb
```diff
@@ -0,0 +1,517 @@
...
+        expect(subject.errors).to be_empty
+      end
+    end
+
+    context "when medical_conditions have unexpected value" do
```
leenagupte:
```suggestion
    context "when medical_conditions has an unexpected value" do
```

spec/services/json_validator_spec.rb
```diff
@@ -0,0 +1,517 @@
...
+      let(:data) { valid_data.merge(medical_conditions: "unexpected") }
+
+      it "returns a medical conditions property error" do
+        expect(error["data"]).to eq("unexpected")
+        expect(error["data_pointer"]).to include("medical_conditions")
+        expect(error["schema_pointer"]).to eq("/properties/medical_conditions")
+      end
+    end
+
+    context "when name missing" do
```
leenagupte:
```suggestion
    context "when name is missing" do
```

spec/services/json_validator_spec.rb
```diff
@@ -0,0 +1,517 @@
...
+      let(:data) { valid_data.except(:name) }
+
+      it "returns a missing keys error" do
+        expect(error["missing_keys"]).to contain_exactly("name")
+      end
+    end
+
+    context "within name" do
```
leenagupte:
```suggestion
    context "#name" do
```

spec/services/json_validator_spec.rb
```diff
@@ -0,0 +1,517 @@
...
+      let(:missing) { :first_name }
+      let(:data) do
+        valid_data.tap do |valid_data|
+          valid_data[:name].delete(missing)
+        end
+      end
+
+      context "when first_name missing" do
```
leenagupte:
```suggestion
      context "when first_name is missing" do
```

spec/services/json_validator_spec.rb
```diff
@@ -0,0 +1,517 @@
...
+        expect(error["missing_keys"]).to contain_exactly("name")
+      end
+    end
+
+    context "within name" do
+      let(:missing) { :first_name }
+      let(:data) do
+        valid_data.tap do |valid_data|
+          valid_data[:name].delete(missing)
+        end
```
leenagupte: I think it would be easier to read if this happened in each test, rather than having to scroll up to read it each time.

spec/services/json_validator_spec.rb
```diff
@@ -0,0 +1,517 @@
...
+      end
+
+      context "when first_name missing" do
+        it "returns a missing keys error" do
+          expect(error["missing_keys"]).to contain_exactly("first_name")
+        end
+      end
+
+      context "when middle_name missing" do
```
leenagupte:
```suggestion
      context "when middle_name is missing" do
```

spec/services/json_validator_spec.rb
```diff
@@ -0,0 +1,517 @@
...
+        let(:missing) { :middle_name }
+
+        it "returns no error" do
```
leenagupte:
```suggestion
        it "returns no errors" do
```

spec/services/json_validator_spec.rb
```diff
@@ -0,0 +1,517 @@
...
+      let(:data) { valid_data.except(:live_in_england) }
+
+      it "returns a missing keys error" do
+        expect(error["missing_keys"]).to contain_exactly("live_in_england")
+        expect(error["data"]).to be_nil
+      end
+    end
+
+    context "when live_in_england in valid" do
+      let(:data) { valid_data.merge(live_in_england: "invalid") }
```
leenagupte:
```suggestion
      let(:invalid_data) { valid_data.merge(live_in_england: "invalid") }
```

I think "data" is an overloaded term as it's also something that is returned in `errors`

spec/services/json_validator_spec.rb
```diff
@@ -0,0 +1,517 @@
...
+    it "returns no error if the data is valid" do
+      expect(subject.errors).to be_empty
+    end
+
+    context "when live_in_england missing" do
+      let(:data) { valid_data.except(:live_in_england) }
+
+      it "returns a missing keys error" do
+        expect(error["missing_keys"]).to contain_exactly("live_in_england")
+        expect(error["data"]).to be_nil
```
leenagupte, line 40: Is there a test for when `error["data"]` is not `nil`?

spec/services/json_validator_spec.rb
```diff
@@ -0,0 +1,517 @@
...
+
+      context "when middle_name missing" do
+        let(:missing) { :middle_name }
+
+        it "returns no error" do
+          expect(subject.errors).to be_empty
+        end
+      end
+
+      context "when last_name missing" do
```
leenagupte:
```suggestion
      context "when last_name is missing" do
```

app/services/json_validator.rb
```diff
@@ -0,0 +1,38 @@
...
+        error["details"]
+      elsif error["data_pointer"].present?
+        error.slice("data", "data_pointer", "schema_pointer")
+      else
+        {
+          type: "Unknown error",
+          data_keys: data.keys,
+          error_keys: errors.keys,
+        }
+      end
```
leenagupte: The tests check that error["missing_keys"] has the correct value, but there's no reference to "missing_keys" in the code. Where is it coming from?

spec/services/json_validator_spec.rb
```diff
@@ -0,0 +1,517 @@
...
+        expect(error["missing_keys"]).to contain_exactly("live_in_england")
+        expect(error["data"]).to be_nil
+      end
+    end
+
+    context "when live_in_england in valid" do
+      let(:data) { valid_data.merge(live_in_england: "invalid") }
+
+      it "returns a yes_no definition error against field" do
+        expect(error["data"]).to eq("invalid")
```
leenagupte: Is `error["data"]` either `nil` or `"invalid"`?

spec/services/json_validator_spec.rb
leenagupte:
```suggestion
      context "when middle_name is missing" do
```

spec/services/json_validator_spec.rb
leenagupte:
```suggestion
        it "returns no errors" do
```

spec/services/json_validator_spec.rb
leenagupte:
```suggestion
      context "when last_name is missing" do
```

spec/services/json_validator_spec.rb
```diff
@@ -0,0 +1,517 @@
...
+      context "when last_name missing" do
+        let(:missing) { :last_name }
+
+        it "returns a missing keys error" do
+          expect(error["missing_keys"]).to contain_exactly("last_name")
+        end
+      end
+    end
+
+    context "when date_of_birth missing" do
```
leenagupte:
```suggestion
    context "when date_of_birth is missing" do
```

spec/services/json_validator_spec.rb
```diff
@@ -0,0 +1,517 @@
...
+      let(:data) { valid_data.except(:date_of_birth) }
+
+      it "returns a missing keys error" do
+        expect(error["missing_keys"]).to contain_exactly("date_of_birth")
+      end
+    end
+
+    context "within date_of_birth" do
```
leenagupte:
```suggestion
    context "date_of_birth" do
```

spec/services/json_validator_spec.rb
```diff
@@ -0,0 +1,517 @@
...
+      let(:missing) { :year }
+      let(:replace) { nil }
+      let(:data) do
+        valid_data.tap do |valid_data|
+          valid_data[:date_of_birth].delete(missing) if missing
+          valid_data[:date_of_birth].merge! replace if replace
+        end
+      end
```
leenagupte: This is very confusing to read. I think this set up should be added to each individual test so it is clear exactly what is being tested.

spec/services/json_validator_spec.rb
```diff
@@ -0,0 +1,517 @@
...
+
+      context "when year missing" do
```
leenagupte:
```suggestion
      context "when year is missing" do
```

spec/services/json_validator_spec.rb
```diff
@@ -0,0 +1,517 @@
...
+        it "returns a missing keys error" do
+          expect(error["missing_keys"]).to contain_exactly("year")
+        end
+      end
+
+      context "when year not a number" do
```
leenagupte:
```suggestion
      context "when year is not a number" do
```

spec/services/json_validator_spec.rb
```diff
@@ -0,0 +1,517 @@
...
+        let(:missing) { nil }
+        let(:replace) { { year: "Two Thousand" } }
```
leenagupte: I've forgotten what "replace" is.

spec/services/json_validator_spec.rb
```diff
@@ -0,0 +1,517 @@
...
+
+        it "returns a property error" do
+          expect(error["data"]).to eq("Two Thousand")
+          expect(error["data_pointer"]).to eq("/date_of_birth/year")
+          expect(error["schema_pointer"]).to eq("/properties/date_of_birth/properties/year")
+        end
+      end
+
+      context "when month missing" do
```
leenagupte:
```suggestion
      context "when month is missing" do
```

leenagupte (commented):
Thanks for adding the schema validation tests back, as well as adding why the json-schemer gem is being used to the commit message.

About the schema validation tests. Whilst it's great that they are back, they have become very difficult to read. I've gotten about half-way through reviewing the tests, but the comments I've added to the first set of tests apply to the whole:

The set up for the tests is in before blocks or "lets" that are quite far from the actual tests, so by the time you come to read the test, you've forgotten what they're testing. 

I think the tests could be simplified quite a lot if rather than having nested contexts, there was only one context/describe for each field, and each test it contained did the all of the "Arrange, Act and Assert" it needs. This would have the added bonus of making them easy to read.

Lastly, the "describe", "context" and "it" blocks together should read as a sentence. I've corrected the first set of tests to do this, but it applies to all of the others in the file. 

I think I've added inline comments about it, but I think that the `errors` method in the JsonValidator could be clearer. For example, we test the contents of "missing_keys" which isn't in the code, but I don't think we clearly test `error["details"]`, which is. I think it's worth modelling the errors that we receive back from the validator so that we're only passing back useful information to the caller.

Unfortunately I'm not around to talk about these changes in person. I'm happy for the changes in the first commit to go out, so feel free to separate them out, as they feel separate from the json validation changes.

app/controllers/coronavirus_form/check_answers_controller.rb
reggieb: I've tended to find people don't like `unless` so used a form that matched `if` cleanly. Happy to go for `unless` pattern.

app/services/json_validator.rb
reggieb: It comes from the JSONSchemer error response. `error["details"]` contains a hash `{ "missing_keys" => ["name of key"] }` when a key is missing.

To be honest, the error behaviour of the gem isn't well documented and has inconsistencies. For example when "details" are present, "data" contains most if not all of the submitted data. Whereas, in other errors, "data" holds a very precise output detailing which data key has failed. So in one condition "data" output needs to be returned in the other it needs to be excluded. 

That's part of the reason I thought a comprehensive set of specs were needed to make sure the error behaviours were covered. I took the specs a little further than the originals testing more variations of the pattern matching to ensure all likely error behaviours were tested.

It's also why I left a catch all in the `else` block as there may be (though I think not) an error condition that I've missed.

spec/services/json_validator_spec.rb
reggieb: It's testing a different method. `valid` and `errors` are two different methods, using two different calls on the gem. `valid` returns a simple boolean, where as `validate` returns an errors hash. So yes it's the same invalid data structure - it's a known invalid data structure.

spec/services/json_validator_spec.rb
reggieb: Yes - there are a number of tests that look for `error["data"]` content. I added this one specially because the data content returned by JSONSchemer's `validate` method in this condition contains all the data submitted - so it important that it doesn't get logged. I'll split that out into a separate spec to make that clearer.

spec/services/json_validator_spec.rb
reggieb: I've renamed it `json_data`. I was tempted to call it `json` but as it's actually a Hash, added the `_data` suffix.

spec/services/json_validator_spec.rb
reggieb: Name isn't an instance method, it's a key within a hash. So `#name` is incorrect. I'll change it to `'within "name"'`

spec/services/json_validator_spec.rb
reggieb: `error["data"] contains the string passed in that is causing the failure. So here, the string "invalid".

It's an idiom I've seen a lot, that when you need a string that behaves a certain way you use the description of that behaviour as the string. So for an unknown string use "unknown" for example. However, as that seems to be confusing here, I'll use "foo" as used in the old specs.

reggieb:
@leenagupte thank you for your review.
- I've updated the texts to be more readable as you suggested.
- I've also refactored the specs so that they each contain a more complete construction of the json data
- I've rename `data` to `json_data` so that it isn't confused with the "data" key returned from JSON::Validator.
- I've added a comment about how the errors returned from JSON::Validator behave as it isn't obvious and has taken me a little time to work out.

I've kept the `context`, `let`, `it` structure as that's [good practice for RSpec specs](http://www.betterspecs.org/), but I hope that by bringing the `json_data` construction closer to each `it` clause I've addresses your issue with having to skip back and forth through the spec to work out what's happening.

steveholmes:
Nice one! I believe all of Leena's comment are addressed in the updates. (I had earlier approved an earlier version so not able to approve twice :) But :+1: from me) 
Needs a rebase as there are conflicts but otherwise looks good to go.

PR #455, opened 2020-07-22
Reduce number of PaaS instances

JonathanHallam:
This will reduce the number of PaaS instances in production from 20 to 2. This is because the service has been shut down. 2 instances will be enough to keep the redirect working.

Trello - https://trello.com/c/vYHAxDKJ/706-reduce-number-of-paas-instances-for-ev-form

### DIFF
--- a/concourse/pipeline.yml
+++ b/concourse/pipeline.yml
@@ -173,7 +173,7 @@ jobs:
         file: git-master/concourse/tasks/deploy-to-govuk-paas.yml
         params:
           CF_SPACE: production
-          INSTANCES: 20
+          INSTANCES: 2
           WORKER_INSTANCES: 1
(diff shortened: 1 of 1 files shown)


gclssvglx (approved):
LGTM 👍

issyl0:
Line 111, I guess we can reduce staging's instances from 5 to 2 as well?

JonathanHallam:
> Line 111, I guess we can reduce staging's instances from 5 to 2 as well?

I'd rather not for the time being, simply because we're going to be keeping staging running for now. It's only being publicly halted. There's a chance we'll need to work on it in the background.

Rosa-Fox:
@JonathanHallam [Smoke test removing PR](https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/pull/456) has been deployed so free feel to 🚢

PR #465, opened 2020-08-13
Update privacy policy

StephenGill:
Update privacy policy to reflect changes following new registrations being stopped: https://trello.com/c/nRqD3g2M/244-re-publish-wave-1-privacy-policy

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -106,7 +106,7 @@ en:
   privacy_page:
     title: Privacy
     body_text: |
-      <p class="govuk-body">Last updated: 30 April 2020</p>
+      <p class="govuk-body">Last updated: 14 August 2020</p>
       <p class="govuk-body">This service is run by the Cabinet Office, working with:</p>
       <ul class="govuk-list govuk-list--bullet">
         <li>the Ministry of Housing, Communities & Local Government</li>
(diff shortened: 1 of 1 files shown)


config/locales/en.yml
```diff
@@ -106,57 +106,73 @@ en:
...
+        <li>name</li>
+        <li>address</li>
+        <li>date of birth</li>
+        <li>email address</li>
+        <li>phone numbers</li>
+        <li>NHS number</li>
+        <li>whether the person needs help getting food and essential supplies</li>
+        <li>whether the person needs help meeting their basic care needs</li>
+        <li>whether the person has special dietary requirements</li>
+        <li>Nwhether there’s someone in the person’s household who can carry a box of supplies inside</li>
```
JonathanHallam: don't think this N is meant to be there

StephenGill: Sorry, yes - it's a typo. Is it easier for you to fix it, or would you rather I did it?

JonathanHallam: I'll fix it now. There also seems to be something that's caused this not to work but I'm not sure what it is yet

issyl0 (approved):
Looks like some HTML in some YAML.

